PR #11, opened 2017-11-09
added missing instantiation

fiblan:
just an oversight


mcollina:
Thanks for catching this!

Issue #29, opened 2018-02-12
Socket closed by remote peer


alizahid:
Had this problem since I upgraded to `0.4.0`.

When I upload a file, the socket is immediately closed by the server. The file upload happens in the background.

I need to wait for the upload to complete, and then whatever processing to be done, so I can return the final url from AWS. I thought I was doing something wrong, but I just tried the sample code in a new project and it has the same behavior.

mcollina:
Can you provide steps to reproduce?

mcollina:
This is already out in 0.4.1.

alizahid:
Ah. I was watching the repo for a tag release. Thank you so much! 🍻

PR #30, opened 2018-02-14
do not abort the outgoing response

mcollina:
Fixes https://github.com/fastify/fastify-multipart/issues/29

@alizahid can you check if this fixes your issue?


alizahid:
@mcollina It works! 🎩

PR #34, opened 2018-04-02
Expose busboy options, update docs and tests.

nigelhanlon:
This PR allows end users to pass options to the internal busboy module when registering the plugin with fastify. This is useful as you currently cannot set upload limits which are supported by busboy internally. 

The supported options are documented on the [busboy page here](https://github.com/mscdex/busboy#busboy-methods). I have included an updated readme and updated tests to reflect the proposed changes. Feed back and suggestions welcome.


index.js
```diff
@@ -40,7 +40,7 @@ function fastifyMultipart (fastify, options, done) {
 
     const req = this.req
 
-    const stream = new Busboy({ headers: req.headers })
+    const stream = new Busboy(Object.assign({ headers: req.headers }, options))
```
mcollina: @nigelhanlon I would prefer if we copied the options using an `Object.keys()` and `for`  loop instead. `Object.assign()`  is super slow and it will likely still be slow in Node 10 (see https://twitter.com/dan_abramov/status/980436488860196864).

index.js
```diff
@@ -40,7 +40,11 @@ function fastifyMultipart (fastify, options, done) {
 
     const req = this.req
 
-    const stream = new Busboy(Object.assign({ headers: req.headers }, options))
+    const busboyOptions = { headers: req.headers }
+    for (let opt of Object.keys(options)) {
+      busboyOptions[opt] = options[opt]
+    }
```
mcollina: `for of` is super slow as well (also using `let` in loops is slow), you'd need a nice old `for (var i = 0; i< n; i++)`.

nigelhanlon: Do you mean something like this? 

```js
    const busboyOptions = { headers: req.headers }
    const keys = Object.keys(options)
    for (var i = 0; i < keys.length; i++) {
      busboyOptions[keys[i]] = options[keys[i]]
    }
    const stream = new Busboy(busboyOptions)
```

mcollina:
Can you remove the spurious package-lock file and add it to .gitignore?

PR #56, opened 2019-02-04
feat: add typings to org repo

jannikkeye:
Motivation: [see here](https://github.com/fastify/fastify-static/issues/66#issuecomment-460159590)

This PR adds a modified version of my typings on DefinitelyTyped to the org repo as that seems to be the consensus now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,18 @@
 language: node_js
+
 node_js:
   - "8"
   - "6"
   - "10"
+  - "11"
+
+os:
+  - linux
+  - windows
+
+script:
+  - npm run lint && npm run coveralls && npm run typescript
+
 notifications:
   email:
     on_success: never
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ os:
   - windows
 
 script:
-  - npm run lint && npm run coveralls && npm run typescript
+  - npm run lint && npm run typescript
 
 notifications:
   email:
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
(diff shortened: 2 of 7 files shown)


jannikkeye:
I forgot to update the unit test script here, but somehow the tests are failing for me, locally anyway. I tried resetting to a specific commit and one test also fails there. I'm on MacOS and tried different node versions to no success. I tried on commit: 3bb2b07b2fcec9b29d83cd9f22e4cf7c880138f8

.travis.yml
```diff
@@ -1,8 +1,18 @@
...
   - "6"
   - "10"
+  - "11"
+
+os:
+  - linux
+  - windows
+
+script:
+  - npm run lint && npm run coveralls && npm run typescript
```
mcollina: `npm run coveralls` does not exists.

mcollina:
One test is a bit flaky, don't worry about it.

.travis.yml
jannikkeye: Indeed it does not. Oversight on my part, carryover from another plugin repo.

jannikkeye:
I cannot commit my changes because that one test keeps failing.

```
npm run unit

> fastify-multipart@0.6.0 unit /Users/janisis/code/fastify-multipart
> tap test/*.test.js

test/multipart.test.js .............................. 42/43
  should call finished when both files are pumped
  not ok should be equal
    +++ found                                                           
    --- wanted                                                          
    -2                                                                  
    +0                                                                  
    compare: ===
    at:
      line: 97
      column: 9
      file: test/multipart.test.js
      type: Immediate
      method: _onImmediate
    stack: |
      Immediate._onImmediate (test/multipart.test.js:97:9)
    source: |
      t.equal(fileCount, 2)

total ............................................... 42/43
  

  42 passing (10s)
  1 failing
```

mcollina:
`git commit -n`

Issue #59, opened 2019-03-08
Schema validation


olivierto:
Hello there! 
I'm trying to get  fastify-multipart form and fastify/ajv schema validation to work together in a restfull api. 
But I still get an error : 
```json 
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "body should be object",
    "validation": [
        {
            "keyword": "type",
            "dataPath": "",
            "schemaPath": "#/type",
            "params": {
                "type": "object"
            },
            "message": "should be object"
        }
    ],
    "code": "BadRequest"
}
``` 
I've extented AJV with a fileType 
```javascript
ajv.addKeyword("isFileType", {
        compile: (schema, parent, it) => {
            // Change the schema type, as this is post validation it doesn't appear to error.
            (parent as any).type = "file";
            delete (parent as any).isFileType;
            return () => true;
        },
    });
```
Here is my schema
```json
{
    "consumes": ["multipart/form-data"],
    "body": {
      "type": "object",
      "properties": {
       "urbanFile": {
          "isFileType": true,
          "type": "object"
        },
        "authorizationType": {
          "type": "string",
          "enum": [
            "buildingPermit",
            "planningPermit",
            "demolitionPermit"
          ]
        },
}
``` 
and here is my handler function 
```javascript
const mutliPart = request.multipart(
      function fileHandler(field, file, filename, encoding, mimetype) {
        console.log("file is in " + field + " and is named " + filename);
      },
      function done(error) {
        if (error) {
          throw error;
        }
      }
    );

    mutliPart.on('field', function (key, value) {
      console.log('form-data', key, value)
    })
    reply.status(201).send("Victory");
```
Is this a bug ? or am I doing something wrong?
any help on this would be much much appreciated. 

Cheers,
Olivier

Eomm:
Hi, I could be faster to help you if you can provide me a complete snippet to replicate the issue because I need to know how you configure the ajv instace and the http request you send (a CURL would be really helful)

olivierto:
thanks for you reply, but after few hours i get this to get working.
Here is my setup fastify
```javascript
const ajv = new Ajv({
        // removeAdditional: "all",
        useDefaults: true,
        coerceTypes: true,
        $data: true,
        extendRefs: true
    });
    require('ajv-keywords')(ajv);

    // Add isFileType for file / multipart-form data validation
    ajv.addKeyword("isFileType", {
        compile: (schema, parent, it) => {
[...]

PR #60, opened 2019-03-14
add feature to append to body the form parameters

Eomm:
This feature implements #59 adding an option to write in the `req.body` the parameters and let add a validation to the parameters.


README.md
```diff
@@ -105,6 +105,68 @@ fastify.post('/', function (req, reply) {
...
+        myStringField: { type: 'string' },
+        myFilenameField: {
+          type: 'object',
+          properties: {
+            encoding: { type: 'string' },
+            filename: { type: 'string' },
+            limit: { type: 'boolean' },
+            mimetype: { type: 'string' }
+          }
+        }
```
mcollina: I'm not convinced by the fact that these properties are defined by the plugin and so they will be fixed for the user. Maybe we can use a shared schema that is added by the plugin? 

These will also show up in swagger - and this is really not ideal. I don't now how to handle these.

Eomm: > Maybe we can use a shared schema that is added by the plugin?

This could be helpful! So the user must set only a set of key-value:
```js
{
  myStringField: { type: 'string' },
  myFilenameField: 'fileType#',
  other: { type: 'number' },
  ...
}
```

> These will also show up in swagger - and this is really not ideal. I don't now how to handle these.

In the first place we can hide the file properties: now we are not supporting the `type: file`, so I think that when we will support it we can fix this plugin.
I'm working on the swagger side too but I wouldn't add to this PR also those considerations.
In `fastify-swagger` the work will take much longer

mcollina: Let's do the shared schema then, something like "MultipartFileType" or something that we can use to avoid collision with a user type.

Eomm: I think I have done 👍

index.js
```diff
@@ -12,7 +12,63 @@ function setMultipart (req, done) {
...
+      limit: false
+    }
+
+    const fileData = []
+    const isCustomDataConsumer = typeof options.onData === 'function'
+    const onData = isCustomDataConsumer
+      ? options.onData
+      : (fieldName, data) => { fileData.push(data) }
+
+    file.on('data', (data) => { onData(field, data) })
```
mcollina: I’m not convinced we should do this. In this way we do not support backpressure at all: it is going to cause issues. I think we should just remove the onData option.

index.js
```diff
@@ -12,7 +12,63 @@ function setMultipart (req, done) {
...
+    file.on('limit', () => { body[field].limit = true })
+
+    file.on('end', () => {
+      if (!isCustomDataConsumer) {
+        body[field].data = Buffer.concat(fileData.map((part) => Buffer.isBuffer(part) ? part : Buffer.from(part)))
```
mcollina: The type is not really going to change. If you are reading buffers, you’ll always read buffers. If you are concatenating strings, you should just use +.

mcollina: you can just use concat-stream

index.js
Eomm: Then I don't understand how we could consume the stream: we put the chunk in an array as was the default? Or are you thinking another way?
Maybe the user could pass as argument a Stream Writer Factory where we will pipe the input stream?

test/multipart.test.js
```diff
@@ -319,3 +319,303 @@ if (!process.env.TRAVIS) {
     })
   })
 }
+
+test('append to body option', t => {
```
mcollina: All the new tests are probably better off in a new file.

index.js
mcollina: `onData` seems to cater for the advanced use case (otherwise I would keep the default), but **`onData` does not support backpressure**:  an advanced user will not be able to use it. If you want to `attachToBody`, we can assume you want to take the whole file in memory.

One way to handle this use case is to have an `onFile` callback, where a stream is passed in. An implementer would have to guarantee that the stream is fully consumed.

> Then I don't understand how we could consume the stream: we put the chunk in an array as was the default? 

Yes essentially.

Eomm:
Fixed the comments 👍🏼

index.js
```diff
@@ -12,7 +12,65 @@ function setMultipart (req, done) {
   done()
 }
 
+function attachToBody (options, req, reply, next) {
+  req[kMultipart] = true
+
+  const userWantConsumeStream = typeof options.manageUploadStream === 'function'
+  const consumerStream = options.manageUploadStream
```
mcollina: I think we should implement the accumulating functionality as a default implementation of this function.

Eomm: Yeah, now it is more clear to read

index.js
```diff
@@ -12,7 +12,58 @@ function setMultipart (req, done) {
   done()
 }
 
+function attachToBody (options, req, reply, next) {
+  req[kMultipart] = true
+
+  const defaultConsumer = (field, file, filename, encoding, mimetype) => {
```
mcollina: Can you make this a top level function? I think we should also pass in the body object.

index.js
```diff
@@ -12,7 +12,58 @@ function setMultipart (req, done) {
...
+    body[field] = {
+      data: [],
+      filename,
+      encoding,
+      mimetype,
+      limit: false
+    }
+    consumerStream(field, file, filename, encoding, mimetype)
+  }, function (err) {
+    req.body = body
```
mcollina: Is this correct? What’s your reasoning?

index.js
Eomm: I leave the body out from the parameter (and used the clojure instead) because I was thinking that could be dangerous pass also the body object because the user could manipulate it in a wrong pattern (without using hooks for example).

Am I too careful?

index.js
Eomm: After composing the `body` object I will attach it to the request body.
This is the function executed on `done` that could and could not have the error.
Should I check the error before adding the body to the request?

mcollina: I think so, yes.

index.js
mcollina: The problem with that approach is that a user cannot implement this function with the public API. They can only give in a top-level function, while you are creating a closure here. The functionality should be identical.

A simplest thing would be to expose the body.

index.js
```diff
@@ -12,7 +12,58 @@ function setMultipart (req, done) {
...
+  const body = { }
+  const mp = req.multipart((field, file, filename, encoding, mimetype) => {
+    body[field] = {
+      data: [],
+      filename,
+      encoding,
+      mimetype,
+      limit: false
+    }
+    consumerStream(field, file, filename, encoding, mimetype)
```
mcollina: You might want to check if this returns a promise and attach a .catch() handler.

README.md
```diff
@@ -105,6 +105,80 @@ fastify.post('/', function (req, reply) {
 })
 ```
 
+You can also use all the parsed HTTP request parametes to the body:
+
+```js
+const options = {
+  addToBody: true,
+  sharedSchemaId: 'MultipartFileType', // Optional shared schema id
+  onFile: (fieldName, stream, filename, encoding, mimetype) => {
```
mcollina: the body arg is missing in the docs

Eomm:
May I release it as minor? (now the version is `0.7.0`) 

> LGTM, good work!

Thank you for the patience 👍

mcollina:
minor it is

Il giorno gio 4 apr 2019 alle 20:00 Manuel Spigolon <
notifications@github.com> ha scritto:

Eomm:
I can't do it 😅

Looking the repositories list of my team, it seems that [this repo is not in](https://github.com/orgs/fastify/teams/fastify/repositories?before=Y3Vyc29yOnYyOpKuZmFzdGlmeS1zdGF0aWPOBf7B_g%3D%3D)

There are 36 repo, some other is missing

mcollina:
now you can ;)

Issue #64, opened 2019-04-17
Multiple files upload with addToBody stores only one file


SkeLLLa:
Hi. As far as can see from the code and examples if multiple files will be uploaded with singe file input, it will put them into the same field in body and `filename` and other fields inside it will correspond to one of the files.

Multiple files could be tracked by number of calls of `defaultConsumer` or if that function is called with different filename. So I think it's better to track it by call number which will represent an index in array. According to this field type in body could be also an array of items with `sharedSchemaId` type. And if that array has only one file, then, probably, it could replace array, so we'll get an file object as it's done right now.

mcollina:
Definitely, would you like to send a PR?

SkeLLLa:
Yes I can, probably in couple of days if someone will not make it faster.

But I have a question regarding filed type. If we have input with name `file` should the key `file` in body to be an array even if there's only one file in there. Or should we put an object instead? E.g.

```
body = {
  file: [{name, data, ...etc}]
}
```
or
```
body = {
  file: {name, data, ...etc}
}
```

Also I'm thinking about an idea of changing the default behaviour of this module. I think it would be better if this module will use `addToBody` behaviour as default one. It will be easier to use it and it will cover most of user usecases. Also it will make code look prettier, because all the stuff with event listeners, onFinish handlers will be done under the hood.

And if someone will need "streams API" I think it can be done in the same manner. But in field `file` for example we'll have and array of object where data will be a readable stream instead of buffer. And we'll add, for example, extra filed to request object that will be a promise, which can be awaited to ensure that all streams were read.

That's just some thoughts, that I think would be nice to have to release version 1.0.0.
WDYT, @mcollina.

SkeLLLa:
I've made some progress regarding the PR, but I'm stuck in writing tests, because it seems that `form-data` doesn't support multiple values in same field, like browser does.

nwoltman:
@SkeLLLa In what way does it not support multiple values? Appending to the same field should work.

```js
form.append('upload', file1)
form.append('upload', file2)
```

SkeLLLa:
@nwoltman I've tried such approach, but in that way I receive only `file2` in upload field. Maybe the bug is somewhere in my changes, but when I use html form from browser all is working fine.

Eomm:
> should the key file in body to be an array even if there's only one file in there. Or should we put an object instead?

I think that it should if we don't want to release a semver-major change.
[...]

PR #66, opened 2019-04-19
Fix/multiple files in body

SkeLLLa:
This PR includes changes from https://github.com/fastify/fastify-multipart/pull/63 so it should be merged first.

It currently makes behaviour slightly different, so file field will be an array by default, even if there's only one file. Please let me know if it's ok or not. And after that I'll update documentation accordingly.


Eomm:
This PR closes #64 

I would like only to remind that, when we will define how to manage the behaviour of this fix, a note in the readme is needed 👍

SkeLLLa:
@Eomm, yes, that's planned if others confirm that receiving array in any cases of files is ok.

index.js
```diff
@@ -13,20 +13,27 @@ function setMultipart (req, done) {
...
+  }
 
   const consumerStream = options.onFile || defaultConsumer
   const body = { }
+  const fileCount = { }
   const mp = req.multipart((field, file, filename, encoding, mimetype) => {
-    body[field] = {
+    fileCount[field] = fileCount[field] || 0
+    body[field] = body[field] || []
+    body[field][fileCount[field]] = {
```
mcollina: I don’t understand this fileCount logic. I think only one of these it’s called at any given time, so we can just use .push.

index.js
```diff
@@ -46,12 +53,12 @@ function attachToBody (options, req, reply, next) {
   })
 }
 
-function defaultConsumer (field, file, filename, encoding, mimetype, body) {
+function defaultConsumer (field, file, filename, encoding, mimetype, body, fileCount) {
```
nwoltman: `fileCount` should be called `fileIndex` here.

index.js
SkeLLLa: That just was first solution came in mind. I've replaced it with push, tests seems fine.

index.js
SkeLLLa: done

index.js
```diff
@@ -46,12 +50,12 @@ function attachToBody (options, req, reply, next) {
   })
 }
 
-function defaultConsumer (field, file, filename, encoding, mimetype, body) {
+function defaultConsumer (field, file, filename, encoding, mimetype, body, fileIndex) {
   const fileData = []
   file.on('data', data => { fileData.push(data) })
-  file.on('limit', () => { body[field].limit = true })
+  file.on('limit', () => { body[field][fileIndex].limit = true })
```
mcollina: There should be no need for `fileIndex`. It's always the last file in the list.

SkeLLLa: Do you mean that we should define it inside function instead of passing as an argument? Because in this particular line I'm not sure that it will be the last file in array. As well as in next line.

mcollina: It it always the last file. The full stream must be consumed for the next one to start.

SkeLLLa: Got it. Thanks.

test/append-body.test.js
```diff
@@ -489,3 +556,25 @@ test('append to body without files and shared schema', t => {
     })
   })
 })
+
+test('append to body option does not change behaviour on not-multipart request', t => {
```
mcollina:
```suggestion
test('addToBody option does not change behaviour on not-multipart request', t => {
```

test/append-body.test.js
```diff
@@ -140,6 +140,73 @@ test('append to body option and multiple files', t => {
   })
 })
 
+test('append to body option and multiple files in one field', t => {
```
mcollina:
```suggestion
test('addToBody option and multiple files in one field', t => {
```

Can you also update this in the other tests? Thanks.

SkeLLLa: @mcollina you mean change that in all tests, even those that wasn't changed, right?

mcollina: yes

SkeLLLa: done

.gitignore
```diff
@@ -38,3 +38,4 @@ jspm_packages
 
 # Ignore package-lock
 package-lock.json
+.vscode
```
nwoltman: Please ensure this file ends with a newline.

README.md
```diff
@@ -146,8 +146,10 @@ fastify.post('/', function (req, reply) {
 The options `onFile` and `sharedSchemaId` will be used only when `addToBody: true`.
 
 The `onFile` option define how the file streams are managed:
-+ if you don't set it the `req.body.<fieldName>.data` will be a Buffer with the data loaded in memory
-+ if you set it with a function you **must** consume the stream and the an the `req.body.<fieldName>.data` will be an empty array
++ if you don't set it the `req.body.<fieldName>[index].data` will be a Buffer with the data loaded in memory
++ if you set it with a function you **must** consume the stream and the an the `req.body.<fieldName>[index].data` will be an empty array
```
nwoltman:
```suggestion
+ if you set it with a function you **must** consume the stream, and the `req.body.<fieldName>[index].data` will be an empty array
```

.gitignore
SkeLLLa: done

PR #68, opened 2019-05-09
add body to onFile in type definition file

pikanezi:
The body parameter was missing in the onFile type.

I set the type as object since I was not sure what to use, if you have a better idea, let me know.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -35,7 +35,7 @@ declare const fastifyMultipart: fastify.Plugin<Server, IncomingMessage, ServerRe
     /**
      * Manage the file stream like you need
      */
-    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: object) => void;
+    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: any) => void;
 
     limits?: {
         /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -35,7 +35,7 @@ declare const fastifyMultipart: fastify.Plugin<Server, IncomingMessage, ServerRe
     /**
      * Manage the file stream like you need
      */
-    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string) => void;
+    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: object) => void;
 
     limits?: {
         /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,7 @@ import busboy = require("busboy");
 import fastify = require("fastify");
 
 import { Server, IncomingMessage, ServerResponse } from 'http';
-import { Readable } from "stream";
+import { Readable } from 'stream';
 
 type MultipartHandler = (
     field: string,
@@ -14,7 +14,22 @@ type MultipartHandler = (
     mimetype: string,
 ) => void;
 
-type MultipartOptions = {
+export interface BodyEntry {
+    data: Buffer,
+    filename: string,
+    encoding: string,
+    mimetype: string,
+    limit: false
+}
+    
+declare module "fastify" {
+    interface FastifyRequest<HttpRequest> {
+        isMultipart: () => boolean;
+        multipart: (handler: MultipartHandler, next: (err: Error) => void) => busboy.Busboy;
+    }
+}
+
+declare const fastifyMultipart: fastify.Plugin<Server, IncomingMessage, ServerResponse, {
     /**
      * Append the multipart parameters to the body object
      */
@@ -28,7 +43,7 @@ type MultipartOptions = {
     /**
      * Manage the file stream like you need
      */
-    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: any) => void;
+    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: Record<string, BodyEntry>) => void;
 
     limits?: {
         /**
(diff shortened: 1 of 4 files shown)


Eomm:
Do you think that the test in [test/types/index.ts](https://github.com/fastify/fastify-multipart/blob/master/test/types/index.ts) could be improved?

pikanezi:
I completed just a little the existing test.

I changed my mind on the type of body, as it can be literally anything, I changed it from object to any.

As for the type of stream in the callback onFiles, I had a [false warning](https://github.com/typescript-eslint/typescript-eslint/issues/46) when importing the type Readable, so I used another syntax which compiles without changing the configuration of the project.

Tell me if you want something else or if it's ok with you.

Eomm:
> As for the type of stream in the callback onFiles, I had a false warning when importing the type Readable, so I used another syntax which compiles without changing the configuration of the project.

For this doubt, we could ask help to @Ethan-Arrowood (I promise to learn types 🙏)

Ethan-Arrowood:
That warning will occur if the linter is not set up correctly. For one, `typescript-eslint` has been deprecated in favor of `@typescript-eslint/parser` and the other surrounding libraries. [Heres a link](https://github.com/MatterhornDev/matterhorn-posts/blob/learn-typescript-linting/learn-typescript-linting.md#27-fixing-unused-variable-definition-error-from-type-import) to a linting guide I'm working on if you want to learn how to fix that error

Eomm:
@pikanezi could you like to update the `typescript-eslint-parser` with the suggestions of Ethan?

pikanezi:
I did it, but as everything regarding eslint, it added a lot of packages...

Eomm:
mmm I saw, maybe you could simply copy from https://github.com/fastify/fastify/blob/master/package.json#L90 ?

Because some of those plugins are not compatible with node 6

index.d.ts
```diff
@@ -28,7 +28,7 @@ type MultipartOptions = {
     /**
      * Manage the file stream like you need
      */
-    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string) => void;
+    onFile?: (fieldName: string, stream: Readable, filename: string, encoding: string, mimetype: string, body: any) => void;
```
evanshortiss: Should any be changed to:

```
body: Record<string, BodyEntry>
```

Where `BodyEntry` is modelled on [this](https://github.com/fastify/fastify-multipart/blob/master/index.js#L24-L30), so might be similar to below:

```
export interface BodyEntry {
  data: Buffer,
  filename: string,
  encoding: string,
  mimetype: string,
  limit: false
}
```

pikanezi: Thanks, I'll add it 🙂

pikanezi:
I've tried to copy the package.json done in fastify as suggested but it still does not work. It works in fastify because in the [typescript test](https://github.com/fastify/fastify/blob/master/test/types/index.ts), they're importing the whole pacakge (* as ..) and they're actually using concrete code from these imports.

Regarding standardjs, if we don't use [this](https://github.com/mightyiam/eslint-config-standard-with-typescript) it just won't works (as seen [here](https://github.com/standard/standard/issues/1099))
But using eslint-config-standard-with-typescript will make the test fail because of nodejs 6 support.

I don't know what to do anymore ☹️

index.d.ts
evanshortiss: This might be possible to improve, e.g encoding might be of type `TranscodeEncoding` from . [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v11/buffer.d.ts#L5)etc.

Eomm:
Don't worry 💪 We can find (a learn) a solution 😃
This PR can be used in future as a perfect example on "how to add type tests" in Fastify plugins ✨

I have checked the other plugins we have and I think [this one](https://github.com/fastify/fastify-cors) can be the right starting point (it is small and it works!)
I saw that it uses the old plugin (`typescript-eslint-parser`): now we know why (the new one doesn't support Node 6) so we must use it that one.

So, let's not lose heart!
I'm sure this time is the good one!! 🐱‍🏍

pikanezi:
I've tried cloning fastify-cors and I added on the top of the file:

```
import { Readable } from 'stream'

const a = (stream: Readable) => stream
```

And it throws me the following errors:
image

I think these tests have been written in a way to avoid the problem because they don't actually test the file index.d.ts on the root of the project.

If I add it to the linted files, it throws multiple errors:
image

Eomm:
Ok, I checked:
@pikanezi  could you download [this patch](https://www.dropbox.com/s/bzrb91tig6fahgy/type.patch?dl=0) and execute `git apply type.patch`?
I don't know how to PR your branch because I have already a fastify-multipart fork 😂

The problem here is that there is an open PR on standard that fixes this issue
https://github.com/standard/standard/pull/1101

I think that this patch isn't perfect due this bug, but accomplish the task.

I don't know how to export the `BodyEntry`.. feel free to fix this 😅

> I think these tests have been written in a way to avoid the problem because they don't actually test the file index.d.ts on the root of the project.

This is not true, I have checked 👍

pikanezi:
I applied your patch (manually because for some reason there was an issue when letting git do it).

For the export of BodyEntry and the plugin itself, I had to change the way the export works in the typing file. That means that will break the import of typescript users using the notation

``` import * as fastifyMultipart from 'fastify-multipart' ```

they will have to import it like

``` import fastifyMultipart from 'fastify-multipart' ```

pikanezi:
I applied your patch and fixed the export of BodyEntry.

Since we export multiple things in the file, I had to make a change that would break for typescript users importing fastify-mutlipart that way:

``` import * as fastifyMultipart from 'fastify-multipart' ```

They would have to fix it like:

``` import fastifyMultipart from 'fastify-multipart' ```

Maybe @Ethan-Arrowood has an idea how to prevent this break?

Ethan-Arrowood:
By exporting BodyEntry you won't be able to maintain the existing `import * as` syntax. I had to handle a similar issue in the fastify [core type refactor](https://github.com/fastify/fastify/pull/1569#issuecomment-482644971). Your best bet is to consider this a semver major change.

Eomm:
@mcollina is a semver-major ok?

PR #78, opened 2019-06-23
Fix default export to commonjs export. Fixes #76.

### LINKED ISSUE #76: Wrong typings
The types of this module specify there's a default export (https://github.com/fastify/fastify-multipart/blob/master/index.d.ts#L81) while the export is a commonjs one (https://github.com/fastify/fastify-multipart/blob/master/index.js#L174).

It seems hard to fix in these typings since there are some other stuff exported, and I see `fastify-plugin` kind of tries to do things with the default export, so I'm wondering should it be fixed here or there ?

Marsup:
As discussed in the issue, changing the default export. Also removing the export that isn't compatible with this format.

I also changed the test to make the server start so you can verify that it works by running something like `ts-node test/types/index.ts`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,7 +14,7 @@ type MultipartHandler = (
     mimetype: string,
 ) => void;
 
-export interface BodyEntry {
+interface BodyEntry {
     data: Buffer,
     filename: string,
     encoding: string,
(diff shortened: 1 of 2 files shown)


mcollina:
@Ethan-Arrowood would you mind taking a look?

Eomm:
It is semver-major right?

mcollina:
I don’t think so. The types were just wrong.

PR #83, opened 2019-07-27
Fix crash on bad input

### LINKED ISSUE #82: Sending empty of invalid multipart data crashes Fastify
Hi,

Here's a minimal example based on the addToBody example from the documentation:

```
const fastify = require('fastify')();

fastify.register(require('fastify-multipart'), {
  addToBody: true
})

fastify.post('/', function (req, reply) {
  console.log(req.body);
  reply.code(200).send();
})

fastify.ready()
  .then(() => {

    fastify.inject({
      method: 'post',
      url: '/',
      headers: {
        'content-type': 'multipart/form-data'
      }
    })
    .catch(error => {
      console.error('Error:', error);
    });
  });
```

The error is: `Multipart: Boundary not found`.

I'd be happy to help if you can point me in the right direction.

schantaraud:
Fixes #82 

Solution provided by https://github.com/mscdex/busboy/issues/140#issuecomment-282794446

It prevents Fastify from crashing on bad input and allows the developer to handle busboy constructor exceptions in Fastify's errorHandler.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const Busboy = require('busboy')
 const kMultipart = Symbol('multipart')
 const eos = require('end-of-stream')
 const deepmerge = require('deepmerge')
+const { PassThrough } = require('stream')
 
 function setMultipart (req, done) {
   // nothing to do, it will be done by the Request.multipart object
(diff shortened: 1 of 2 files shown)


mcollina:
Great work!

schantaraud:
@mcollina done

mcollina:
Done

Issue #102, opened 2019-10-29
Don't attach file to body if limit reached


SkeLLLa:
## 🚀 Feature Proposal

Should this plugin using `attachToBody=true` avoid adding data to body in https://github.com/fastify/fastify-multipart/blob/master/index.js#L60?

## Motivation

If limit is reached we're will not receive the whole file anyway and that data might be useless in most cases. However it will eat memory. Also we can remove that data from memory early, just as soon as `limit` event fired. Cause as far as I can see it takes some time after limit is reached and 
`end` event is fired. I think it depends on file size, so if it would be big enough, then it could be potential attack vector that cause out of memory exception event if we have limit, but will store parts of files in memory.

mcollina:
Would you mind to send a PR?

mcollina:
(Good spot!)

SkeLLLa:
Ok, will do.

StarpTech:
This is covered by the new implementation https://github.com/fastify/fastify-multipart/pull/140. In case of an error, an exception or stream error is thrown and the stream is ignored entirely.

StarpTech:
Closing due to inactivity.

PR #105, opened 2019-11-26
#102 - Dont attach file to body if the limit has been reached

Svjard:
I was having issues with the test "should call finished when both files are pumped" locally but had no changes to affect it. Seems the check on filesCount happens before the callback in pump() is getting executed.

#### Checklist

- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,9 +55,7 @@ function defaultConsumer (field, file, filename, encoding, mimetype, body) {
   const fileData = []
   const lastFile = body[field][body[field].length - 1]
   file.on('data', data => { fileData.push(data) })
-  file.on('limit', () => {
-    lastFile.limit = true
-  })
+  file.on('limit', () => { lastFile.limit = true })
   file.on('end', () => {
     if (!lastFile.limit) {
       lastFile.data = Buffer.concat(fileData)
@@ -144,7 +142,7 @@ function fastifyMultipart (fastify, options, done) {
     })
 
     stream.on('finish', function () {
-      log.debug('finished multipart parsing', files)
+      log.debug('finished multipart parsing')
       if (!completed && count === files) {
         completed = true
         setImmediate(done)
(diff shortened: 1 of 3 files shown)


test/append-body.test.js
```diff
@@ -63,6 +63,49 @@ test('addToBody option', t => {
...
+test('addToBody with limit exceeded', t => {
+  t.plan(5)
+
+  const fastify = Fastify()
+  t.tearDown(fastify.close.bind(fastify))
+
+  fastify.register(multipart, { addToBody: true, limits: { fileSize: 1 } })
+
+  fastify.post('/', function (req, reply) {
+    t.type(req.body.myFile[0].limit, true)
```
Eomm:
```suggestion
    t.equals(req.body.myFile[0].limit, true)
```

test/append-body.test.js
```diff
@@ -63,6 +63,49 @@ test('addToBody option', t => {
...
+    t.type(req.body.myFile[0].data, undefined)
```
Eomm:
```suggestion
    t.equals(req.body.myFile[0].data, undefined)
```

index.js
```diff
@@ -57,7 +57,11 @@ function defaultConsumer (field, file, filename, encoding, mimetype, body) {
   file.on('data', data => { fileData.push(data) })
```
Eomm: we could add the same check here in order to avoid memory usage:

```suggestion
  file.on('data', data => { if (!lastFile.limit) { fileData.push(data) } })
```

index.js
```diff
@@ -57,7 +57,11 @@ function defaultConsumer (field, file, filename, encoding, mimetype, body) {
...
   file.on('limit', () => { lastFile.limit = true })
   file.on('end', () => {
-    lastFile.data = Buffer.concat(fileData)
+    if (!lastFile.limit) {
+      lastFile.data = Buffer.concat(fileData)
+    } else {
+      delete lastFile.data
```
Eomm: This should be a little bit faster
```suggestion
      lastFile.data = undefined
```

PR #107, opened 2019-12-08
Update README.md

theliuk:
Added information about the conditions under which the onEnd callback is not called.

#### Checklist

- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ fastify.post('/', function (req, reply) {
     // be careful of permission issues on disk and not overwrite
     // sensitive files that could cause security risks
     
-    // also, consider that if the file stream is not emptied, the 
+    // also, consider that if the file stream is not consumed, the 
     // onEnd callback won't be called
   }
 })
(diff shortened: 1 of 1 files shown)


RafaelGSS:
Good catch! What do you think that explains the "**Note:** in a repro code?

theliuk:
I don't know, I was looking for a good place where to put this catch.
Also, a friend of mine, who's a more careful reader than me, has noticed something similar to this piece of information at line 120.

README.md
```diff
@@ -59,6 +64,9 @@ fastify.listen(3000, err => {
 
 You can also pass optional arguments to busboy when registering with fastify. This is useful for setting limits on the content that can be uploaded. A full list of available options can be found in the [busboy documentation](https://github.com/mscdex/busboy#busboy-methods).
 
+**Note**: if the file stream that is provided to the handler function is not emptied (like in the example above with the usage of pump) the onEnd callback won't be called at the end of the multipart processing.
```
Eomm: I would use this verb since we used it for the "onFile" option

```suggestion
**Note**: if the file stream that is provided to the handler function is not consumed (like in the example above with the usage of pump) the `onEnd` callback won't be called at the end of the multipart processing.
```

PR #121, opened 2020-05-07
Handle string arrays

### LINKED ISSUE #120: Fields arrays not working
## 🐛 Bug Report

Currently, it's impossible to accept multipart requests with array fields. It's a common pattern to duplicate field to pass an array ([example](https://stackoverflow.com/questions/10002821/how-do-i-post-an-array-with-multipart-form-data-encoding)). Owner of busboy [recommends](https://github.com/mscdex/busboy/issues/159#issuecomment-354678039) to check duplicates of keys and create arrays. But [this line](https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/index.js#L54) is always overriding previous value of the field.

## To Reproduce

```ts
const instance: fastify.FastifyInstance<Server, IncomingMessage, ServerResponse> = fastify();
instance.register(fastifyMultipart, { addToBody: true });
instance.route({
  method: ['GET', 'POST', 'PATCH', 'PUT', 'DELETE', 'HEAD', 'OPTIONS'],
  url: '/echo',
  handler: (request, reply) => {
    reply.code(200).send({
      body: request.body,
      headers: request.headers,
      path: request.req.url,
      method: request.req.method,
    });
  },
});

// then request using any client with the following options
{
    url: `http://${host}/echo`,
    method: 'POST',
    multipart: { someField: ['123', '321'] },
  }
```

## Expected behavior

Expected response body:

```js
{
   headers: {...},
   path: '/echo',
   method: 'post',
   body: {
     someField: ['123', '321'],
   },
}
```

Actual:

```js
{
   headers: {...},
   path: '/echo',
   method: 'post',
   body: {
     someField: '321',
   },
}
```

## Your Environment

- *node version*: 12.16.2
- *fastify version*: 2.14.0
- *os*: Windows
- *fastify-multipart version*: 1.0.5

## Suggestion

Maybe change [this line](https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/index.js#L54) to something like this:

```js
if (body[key] === undefined) {
  body[key] = value
} else if (Array.isArray(body[key])) {
  body[key].push(value)
} else {
  body[key] = [body[key], value];
}
```

morigs:
Resolves #120

#### Checklist

- [x] run `npm run test`
- [x] tests are included
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/test/multipart.test.js
+++ b/test/multipart.test.js
@@ -380,7 +380,7 @@ if (!process.env.TRAVIS) {
 }
 
 test('should not allow __proto__', function (t) {
-  t.plan(3)
+  t.plan(5)
 
   const fastify = Fastify()
   t.tearDown(fastify.close.bind(fastify))
--- a/test/multipart.test.js
+++ b/test/multipart.test.js
@@ -380,7 +380,7 @@ if (!process.env.TRAVIS) {
 }
 
 test('should not allow __proto__', function (t) {
-  t.plan(5)
+  t.plan(3)
 
   const fastify = Fastify()
   t.tearDown(fastify.close.bind(fastify))
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ The `onFile` option define how the file streams are managed:
 + if you don't set it the `req.body.<fieldName>[index].data` will be a Buffer with the data loaded in memory
 + if you set it with a function you **must** consume the stream, and the `req.body.<fieldName>[index].data` will be an empty array
 
-**Note**: By default values in fields with files have array type, so if there's only one file uploaded, you can access it via `req.body.<fieldName>[0].data`
+**Note**: By default values in fields with files have array type, so if there's only one file uploaded, you can access it via `req.body.<fieldName>[0].data`. Regular fields become an array only when multiple values are provided.
 
 The `sharedSchemaId` parameter must provide a string ID and a [shared schema](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#adding-a-shared-schema) will be added to your fastify instance so you will be able to apply the validation to your service like this:
 
--- a/index.js
+++ b/index.js
@@ -51,7 +51,13 @@ function attachToBody (options, req, reply, next) {
       mp.destroy(new Error('__proto__ is not allowed as field name'))
       return
     }
-    body[key] = value
+    if (body[key] === undefined) {
+      body[key] = value
+    } else if (Array.isArray(body[key])) {
+      body[key].push(value)
+    } else {
+      body[key] = [body[key], value]
+    }
   })
 }
 
--- a/test/append-body.test.js
+++ b/test/append-body.test.js
@@ -250,6 +250,52 @@ test('addToBody option and multiple files in one field', t => {
   })
 })
 
+test('addToBody option and multiple strings in one field', t => {
+  t.plan(4)
+
+  const fastify = Fastify()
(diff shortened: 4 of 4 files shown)


test/multipart.test.js
```diff
@@ -380,7 +380,7 @@ if (!process.env.TRAVIS) {
 }
 
 test('should not allow __proto__', function (t) {
-  t.plan(5)
+  t.plan(3)
```
mcollina, line 383: why this change was needed?

morigs: this test doesn't pass on my machine because of the incorrect plan. I have counted assertations by hand and counted 3 assertations. Now the test is passing)

Eomm: I counted and I see 5 assertions:

https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/test/multipart.test.js#L391

https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/test/multipart.test.js#L394

https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/test/multipart.test.js#L420

https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/test/multipart.test.js#L423

https://github.com/fastify/fastify-multipart/blob/d298f193dc4e0d0a446fae70789ceeafc05e3211/test/multipart.test.js#L431

Could you retry to run the test on master in your PC?
I'm afraid the changes do not cover some edge cases

morigs: This is my `npm t` output when `t.plan(5)`:

```
test/append-body.test.js ............................ 65/65
test/multipart.test.js .............................. 53/55 30s
  should not allow __proto__
  not ok test count !== plan
    --- wanted
    +++ found
    -5
    +3
    results:
      ok: false
      count: 3
      pass: 3
      fail: 1
      bailout: false
      todo: 0
      skip: 0
      plan:
        start: 1
        end: 5
        skipAll: false
        skipReason: ''
        comment: ''
      failures:
        - tapError: incorrect number of tests

  not ok test count !== plan
    --- wanted
    +++ found
    -1
    +6
    results:
      ok: false
      count: 6
      pass: 6
      fail: 1
      bailout: false
      todo: 0
      skip: 0
      plan:
        start: null
        end: null
        skipAll: false
        skipReason: ''
        comment: ''
      failures:
        - tapError: no plan

total ............................................. 118/120


  118 passing (33s)
  2 failing
```

According to Tap docs `t.pass` is not an assertion, but `t.error` is. Not sure why this is happening on my machine

Eomm: Try to commit the plan 5 (`git commit -nm 'msg'`), I would like to check CI output

jsumners: `t.pass` is an assertion as the docs state it is literally the basis for all success assertions:

https://node-tap.org/docs/api/#tpassmessage
> Emit a passing test point. This method, and `fail()`, are the basic building blocks of all fancier assertions.

morigs: Ok, I see tests are passing on CI now. It's possible to merge this PR now
I think it's a very strange behaviour of TAP. Perhaps it depends on the OS or node version

PR #127, opened 2020-05-31
Fix content type parser format.

navaru:
Update to fastify v3-rc.3 and fix `addContentTypeParser` to newer format.

#### Checklist

- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.1.1",
     "eslint-plugin-typescript": "^0.14.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fastify-multipart",
-  "version": "2.0.0",
```
Eomm, line 3: Please restore the previous value.

Our release tool will update this field automatically (and it will break if we do it manually 😅 )

navaru: Line reverted.

Issue #139, opened 2020-08-04
Improve developer experience


StarpTech:
## 🚀 Feature Proposal

Today, I needed to develop a simple image upload service in fastify and it felt unsatisfying.

## Motivation

Handling multipart form data should be dead simple and intuitive in fastify.

-  Support `async / await` (async stream iterators)
- Zero boilerplate, provide fastify-hooks to do all this stuff with good defaults (like in expressjs/multer `upload.single("uploaded_file")` )
- Delegate busboy failures to fastify (like in expressjs https://github.com/expressjs/multer/blob/6b5fff5feaf740f249b1b2858e5d06009cbd245c/lib/make-middleware.js#L167-L170)

## Example

In https://eggjs.org/en/basics/controller.html#acquiring-the-submitted-files you can easily switch different modes without dealing with too much boilerplate.

```js
// Single file in disk storage mode
app.post("/images", async (req, res) => {
  req.file.filename
  req.file.mimetype
  // async / await
})

// Multiple files in disk storage mode
app.post("/images", async (req, res) => {
  req.files
})

// Single file in stream mode
app.post("/images", async (req, res) => {
   const stream = await ctx.getFileStream();
})

// Multiple files in stream mode
app.post("/images", async (req, res) => {
 const parts = ctx.multipart();
    let part;
    // parts() return a promise
    while ((part = await parts()) != null) {}
})
```

in fastify you need for the same behaviour a lot of boilerplate and you are responsible to handle all cases manually.

```js
fastify.register(require('fastify-multipart'))

fastify.post('/upload', function (req, reply) {
  const mp = req.multipart(handler, function (err) {
    console.log('###2');
    if (err) {
      reply.send(err)
      return
    }
    console.log('upload completed', process.memoryUsage().rss)
    reply.code(200).send()
  })

  mp.on('field', function (key, value) {
    console.log('form-data', key, value)
  })

  function handler (field, file, filename, encoding, mimetype) {
    pump(file, fs.createWriteStream('a-destination'), (err) => {
      console.log('###1');
    })
  }
})
```

Logs:

```
server listening on 3000
###2
upload completed 31477760
###1
```

Additional problems for the users:
- `onEnd` is fired before the file was written.

fox1t:
What about https://www.npmjs.com/package/fastify-multer for the same old multer DX?

StarpTech:
@fox1t great I didn't know it.

StarpTech:
[...]

Issue #109, opened 2019-12-10
Async Await support


delvedor:
It would be great to have Async Await support in this module!
Would you like to work on it? Let us know below!

ns23:
Hi can i work on it  ?

surajrockzz:
Hii can I work on it?

Swizz:
Hey ! 
Is there any progress on this ?

StarpTech:
Async / await support is fully covered in the new implementation #140

StarpTech:
Closed with #140

PR #140, opened 2020-08-09
feat - new promise based api

StarpTech:
This is a preview of the new interface. I appreciate any feedback. Tests suite is incomplete but the features are working already.

Checkout the `examples` to get a better impression or play with the `form.html`

Related: #109, #139 

#### Checklist


jsumners:
I rather prefer the original interface. This strikes me as too inflexible.

StarpTech:
Could you be more specific? What do you miss?

test/big.test.js
```diff
@@ -1,112 +0,0 @@
...
-const http = require('http')
-const stream = require('readable-stream')
-const Readable = stream.Readable
-const Writable = stream.Writable
-const pump = stream.pipeline
-const eos = stream.finished
-const crypto = require('crypto')
-
-// skipping on Github Actions because it takes too long
-test('should upload a big file in constant memory', { skip: process.env.CI }, function (t) {
```
mcollina, line 16: This test is actually important to check if we do not cause a memory leak.

index.js
```diff
@@ -87,122 +32,213 @@ function busboy (options) {
...
 
-    const busboyOptions = deepmerge.all([{ headers: req.headers }, options || {}, opts || {}])
-    const stream = busboy(busboyOptions)
-    var completed = false
-    var files = 0
-    var count = 0
-    var callDoneOnNextEos = false
+    function onField (name, fieldValue, fieldnameTruncated, valueTruncated) {
+      // don't overwrite prototypes
+      if (getDescriptor(Object.prototype, name)) return
```
mcollina, line 122: Should we error for this instead?

index.js
```diff
@@ -87,122 +32,213 @@ function busboy (options) {
...
+      request.removeListener('close', cleanup)
+      bb.removeListener('field', onField)
+      bb.removeListener('file', onFile)
+      bb.removeListener('close', cleanup)
+    }
+
+    return parts
+  }
+
+  async function saveRequestFiles () {
```
mcollina, line 186: I think this should throw if some of the other methods are called first.

test/append-body.test.js
```diff
@@ -1,788 +0,0 @@
...
-const Fastify = require('fastify')
-const multipart = require('..')
-const http = require('http')
-const path = require('path')
-const fs = require('fs')
-const pump = require('pump')
-
-const filePath = path.join(__dirname, '../README.md')
-
-test('addToBody option', t => {
```
mcollina, line 13: the `addToBody` option is needed to support automatic validation of the body itself if a user would like to do so. If we remove this functionality, we should add some equivalent replacement.

jsumners:
> Could you be more specific? What do you miss?

It looks to me like this implementation assumes "file" uploads and always writes the data out to a file on the filesystem. Multipart isn't always about uploading files. In particular, I have a case where I need to look for a part named "json" and read the data attached to it as an object; I have no need of writing this to the filesystem.

test/big.test.js
StarpTech: Don't worry I just remove it for a clean start. I will add this when we will agree on the implementation.

index.js
StarpTech: I agree.

StarpTech:
> It looks to me like this implementation assumes "file" uploads and always writes the data out to a file on the filesystem. Multipart isn't always about uploading files. In particular, I have a case where I need to look for a part named "json" and read the data attached to it as an object; I have no need of writing this to the filesystem.

Oh no, please check out all examples. You can work with just streams or files.

index.js
StarpTech: Good point.

index.js
```diff
@@ -87,122 +32,213 @@ function busboy (options) {
...
 
+  function handleMultipart (opts = {}) {
     if (!this.isMultipart()) {
-      done(new Error('the request is not multipart'))
-      return
+      throw new Error('the request is not multipart')
     }
 
-    const log = this.log
+    const ch = chan()
```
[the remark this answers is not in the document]
StarpTech: I replaced this module with a simple FIFO. Please check.

jsumners:
> It also looks a lot like a new module to me.

At a minimum, I agree with this statement. One reason I think a major framework has recently been deprecated is due to the way it was routinely using semver major to release what amounted to new frameworks under the same name. This PR completely changes the API for this module, and will be a significant hinderance to people using the module. I think that is a worse developer experience than the current API.

StarpTech:
Wow, why so negative? Let focus at first on the PR. We can discuss afterward how we can incorporate that in an fastify ecosystem.

jsumners:
I'm not seeing any negativity in my statement. I only stated that I agree _this_ module should remain as-is.

StarpTech:
> I'm not seeing any negativity in my statement. I only stated that I agree this module should remain as-is.

You made it already clear with your first [comment](https://github.com/fastify/fastify-multipart/pull/140#issuecomment-671115459).

index.js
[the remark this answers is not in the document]
StarpTech: Check again I could simplify it.

index.js
```diff
@@ -87,122 +33,235 @@ function busboy (options) {
...
-    fastify.addHook('preValidation', function (req, reply, next) {
-      attachToBody(options, req, reply, next)
-    })
-  }
+function fastifyMultipart (fastify, options = {}, done) {
   fastify.addContentTypeParser('multipart', setMultipart)
-
-  fastify.decorateRequest('multipart', multipart)
+  fastify.decorateRequest('handleMultipart', handleMultipart)
+  fastify.decorateRequest('multipart', getMultipartIterator)
```
Eomm, line 39: If we leave this prop with as the old interface `const bb = req.multipart(handler, onEnd)` could be a solution @jsumners ?

jsumners: I think so. But it might be difficult to support both interfaces.

StarpTech: The new interface provides the same flexibility: Consuming streams. On top of that, we can work with files directly and support async/await.

IMO it's not worth supporting both interfaces or what do you see that I don't see?

StarpTech:
@mcollina on windows the tests `npm test` will run in a timeout. When I run the tests via `node` they will fail only occasionally.

E.g it will be stuck.

```
$ node test/multipart-stream.test.js 
TAP version 13
# Subtest: should throw fileSize limitation error on small payload
    1..4
    ok 1 - expect truthy value
    ok 2 - should be equal

```

It looks like it's not related to a specific test. On my machine `'should throw fileSize limitation error on small payload'` and `'should not allow __proto__ as file name'` fail regularly. Perhaps you find the bug in the code.

StarpTech:
@mcollina I could fix it 😄

mcollina:
go for a fix! I don't have a clue - the first is likely due to difference in the network stack.

StarpTech:
On windows ci `npm install` does not even start 🤯 On MAC the tests failing. On Linux works everything fine. I need some help.

mcollina:
> On windows ci npm install does not even start

I do not have any idea about that.

---

I'll take a look for Mac.

StarpTech:
This was super confusing. The other ci were canceled due to `fail-fast` matrix strategy. Now I have a `UnhandledPromiseRejectionWarning: Error: write EPIPE` on Windows and MAC but I can't reproduce it locally.

StarpTech:
@mcollina It looks like that the server closes the request on a file stream error but I have no clue why.

StarpTech:
I found the issue. The stream was not drained after I called `reply.send`.

StarpTech:
I'm on vacation next week. Feel free to review the current API. After that, I will try to reintegrate the old one.

fox1t (commented):
I really like the new API. 👍 
I think that makes this plugin more usable without losing any of its strengths.

test/append-body.test.js
StarpTech: JSON Schema was added.

index.js
```diff
@@ -1,11 +1,22 @@
-'use strict'
```
mcollina, line 1: use strict is still needed

index.js
```diff
@@ -200,9 +275,250 @@ function fastifyMultipart (fastify, options, done) {
...
+        buffer () {
+          if (this._buf) {
+            return Promise.resolve(this._buf)
+          }
+          return new Promise((resolve, reject) => {
+            pump(this.file, concat((buf) => {
+              this._buf = buf
+              resolve(this._buf)
+            })).catch(reject)
+          })
```
mcollina, line 395: I would avoid breaking the promise chain for this. Maybe we can use `this.file` async iterator instead?

index.js
```diff
@@ -200,9 +275,250 @@ function fastifyMultipart (fastify, options, done) {
...
+    file.once('limit', () => {
+      const err = new RequestFileTooLargeError()
+
+      if (file.listenerCount('error') > 0) {
+        file.emit('error', err)
+        logger.warn(err)
+      } else {
+        logger.error(err)
+        // ignore next error event
+        file.on('error', () => { })
```
mcollina, line 450: why is this ignored? Maybe we should log it as debug?

index.js
```diff
@@ -200,9 +230,261 @@ function fastifyMultipart (fastify, options, done) {
...
+
+    bb
+      .on('field', onField)
+      .on('file', onFile)
+      .on('close', cleanup)
+      .on('error', onEnd)
+      .on('finish', onEnd)
+
+    bb.on('partsLimit', function () {
+      const err = new Error('Reach parts limit')
```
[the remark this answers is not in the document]
StarpTech: You refer to an old commit. This is already done.

index.js
```diff
@@ -200,9 +230,261 @@ function fastifyMultipart (fastify, options, done) {
...
+      err.code = 'Request_parts_limit'
+      err.status = 413
+      onError(err)
+    })
+
+    bb.on('filesLimit', function () {
+      const err = new Error('Reach files limit')
+      err.code = 'Request_files_limit'
```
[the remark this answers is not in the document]
StarpTech: Same

index.js
StarpTech: I will log it as an error.

index.js
StarpTech: Good point.

mcollina (approved):
lgtm, let's wait for the others anyway.

StarpTech:
> lgtm, let's wait for the others anyway.

@mcollina I will wait for one more. The activity here is extremely low. I'd like to use this module.

README-legacy.md
```diff
@@ -0,0 +1,194 @@
+# Callback-based API
```
mcollina, line 1: Can you rename this file to `callback.md`?

StarpTech: Done.

mcollina:
Can you change the title of this PR?

delvedor (approved):
LGTM

Issue #131, opened 2020-06-22
Is there proper example on how to cancel upload?


yaneony:
Just wondering if there is some proper example on how to cancel upload on some of event, like "partsLimit", "filesLimit", "fieldsLimit", "limit"? Had anyone success on checking and canceling them?!

mcollina:
Possibly with `req.raw.destroy()`? I've actually never tried.

yaneony:
For an example, i would allow users to create some kind of "content", forum thread, or anything like that... User will be able to upload up to 3 files, 10mb each max. So, someone will upload 10 files instead of 3 and some of those files will be larger than 10mb... There should be some possibility to cancel uploads if any of limits meant above ("partsLimit", "filesLimit", "fieldsLimit", "limit") occurred...

mcollina:
I think calling the callback with an error or `reply.send(new Error(‘too big’))` should do the trick.

yaneony:
`req.raw.destroy()` partially do the trick if you put in into 'limit' event, but, the problem is that application still trying to do something and you don't get any reply from your web/api until some timeout... I'm just looking for any good way to stop busboy processing any of "fields" if any of file/field limits has hit.

yaneony:
Answering my own question: 

To prevent firing "finish" event on any processing after you hit "limit", you will have to "unpipe" the Busboy instance from your request.

That code do the trick:
```
    file.on('limit', () => {
      console.log('File size limit reached');
      req.raw.unpipe(mp);
      res.send(new Error('File size limit reached'));
    });
```

mcollina:
That's awesome! It would be handy to add an `mp.abort()` method that does the above.

yaneony:
So, i've digged around and it looks like that's the only possible variant to cancel upload while hitting the limit. Otherwise, if you don't do that, your app will wait until full file size is consumed.

Following, try to set limit to 1mb and try to upload 700mb file. You will see that it will take as long as it would take to upload full 700mb file and will not break after 1mb is done.

To note is, that it will not work with `addToBody: true` together, since there are no `req` and `res` objects in there.

mcollina:
A PR for this would be amazing!

mcollina:
That seems about how I would handle it! 
One of the things we can do is to add handling of those events in `addToBody`.

yaneony:
One small improvement, you have to add header to close the connection or your browser will hang until it times out. 
Here the code: 

[...]

Issue #129, opened 2020-06-08
onFile function is not called when passed as a per-request option 


cbronson:
## 🐛 Bug Report

The onFile function is not called when passed as an option for a single request. I am not sure if this is a bug or a problem with my code. 

## To Reproduce

```js
fastify.get( ‘/’, options,  
        (req, res) => {  
            const mp = request.multipart(handler, onEnd, {  
	            addToBody: true,  
	            sharedSchemaId:  'MultipartFileType',
	            onFile: (fieldName, stream, filename, encoding, mimetype, body) => {  
			        stream.on(’data’, () => {  
						console.log(‘data…’) //<- no output  
					})  
				},  
			})
		})  
```

Value of `req.body.files`:
```
files: [
    {
      data: <Buffer 38 33 30 39 38 30 34 39 33 32 0a>,
      filename: 'test.pdf',
      encoding: '7bit',
      mimetype: 'application/pdf',
      limit: false
    }
  ]
```

## Expected behavior

The above onFile function should output 'data...' for each received chunk as it does when set as an option on the initial registration and req.body.files[index].data should be an empty array rather than a buffer.

The onFile function is called as expected when set as an option in the initial plugin registration:
```js
fastify.register(require('fastify-multipart'), {
	    addToBody: true, 
            sharedSchemaId:  'MultipartFileType',
	    onFile: (fieldName, stream, filename, encoding, mimetype, body) => {  
		        stream.on(’data’, () => {  
					console.log(‘data…’) //<- prints 'data...' for each chunk as expected  
				})  
			},  
		})
    })
```
Expected value of `req.body.files`:
```
files: [
    {
      data: [],
      filename: 'test.pdf',
      encoding: '7bit',
      mimetype: 'application/pdf',
      limit: false
    }
  ]
```

## Your Environment

- *node version*: 13.14
- *fastify version*: 2.14.1
- *fastify-multipart version*: 1.0.6
- *os*: Ubuntu 20.04 LTS

mcollina:
I think neither `addToBody` or `sharedSchemaId` can actually be specified in the `request.multipart`. As you have experienced, they won't work. I think the doc is wrong and needs to be updated (or possibly the code fixed).

Would you mind to send a PR to update the README?

yaneony:
You might check that example  https://github.com/fastify/fastify-multipart/issues/131

58bits:
[...]

PR #141, opened 2020-08-15
Updated README with note on addToBody issues and handlers

58bits:
Updated README to include a note on `addToBody` per handler/request handling and body schema validation, with references to the issues and examples here https://github.com/fastify/fastify-multipart/issues/129 and https://github.com/fastify/fastify-multipart/issues/131

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ fastify.post('/', function (req, reply) {
 
 You can also use all the parsed HTTP request parameters to the body:
 
-**Note**: Options for `addToBody` and `onFile` are NOT currently available on per-request handlers via options, and so will not work if included as above with `const mp = req.multipart(handler, done, options)`. See these two related issues with examples of how to implement per-handler multipart with fields-to-body schema validation and all stream event handlers - https://github.com/fastify/fastify-multipart/issues/131, https://github.com/fastify/fastify-multipart/issues/129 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,6 +124,8 @@ fastify.post('/', function (req, reply) {
 
 You can also use all the parsed HTTP request parameters to the body:
 
+**Note**: Options for `addToBody` and `onFile` are NOT currently available on per-request handlers via options, and so will not work if included as above with `const mp = req.multipart(handler, done, options)`. See these two related issues with examples of how to implement per-handler multipart with fields-to-body schema validation and all stream event handlers - https://github.com/fastify/fastify-multipart/issues/131, https://github.com/fastify/fastify-multipart/issues/129
```
Eomm, line 127:
```suggestion
**Note**: Options for `addToBody` and `onFile` are NOT currently available on per-request handlers via options, and so will not work if included as above with `const mp = req.multipart(handler, done, options)`. See these two related issues with examples of how to implement per-handler multipart with fields-to-body schema validation and all stream event handlers - [#131](https://github.com/fastify/fastify-multipart/issues/131), [#129](https://github.com/fastify/fastify-multipart/issues/129)
```

vinamra28 (approved):
/lgtm

mcollina:
@StarpTech is this still valid?

StarpTech:
@mcollina not for the new promise-based API. You can't configure `onFile` per route it's a global plugin option.
Regarding "canceling" the new implementation will ignore all data as soon as a `limit` error throws but we have no "abort" method.

PR #145, opened 2020-08-26
fix docs examples

StarpTech:
#### Checklist

- [X] documentation is changed or added

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ try {
 }
 ``` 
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
Can we support both APIs at the same time?

StarpTech:
Should work but it will conflict if you add the schema with the same name twice => `addToBody:true` and `attachFieldsToBody:true` and `sharedSchemaId:123` all other parts are encapsulated to the request.

PR #147, opened 2020-08-27
Rename multipartIterator to parts

mcollina:
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,7 +133,7 @@ function fastifyMultipart (fastify, options = {}, done) {
       })
     }
     fastify.addHook('preValidation', async function (req, reply) {
-      for await (const part of req.multipartIterator()) {
+      for await (const part of req.parts()) {
         req.body = part.fields
         if (part.file) {
           if (options.onFile) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -164,7 +164,12 @@ function fastifyMultipart (fastify, options = {}, done) {
 
   fastify.addContentTypeParser('multipart', setMultipart)
   fastify.decorateRequest(kMultipartHandler, handleMultipart)
+
+  fastify.decorateRequest('parts', getMultipartIterator)
+  // keeping multipartIterator to avoid bumping a major
```
kibertoad, line 169: wonder if this could be done by using some kind of getter that would be emitting warnings when used. although that might require support on fastify level

kibertoad: maybe just a wrapping function would suffice here, actually

mcollina: This has been released for so little that I don't think we should bother with a full blown deprecation.

StarpTech:
I'm not convinced about a rename at this stage but I'm also not against it.

StarpTech (approved):

test/multipart.test.js
```diff
@@ -136,7 +136,7 @@ test('should group parts with the same name to an array', function (t) {
   fastify.register(multipart)
 
   fastify.post('/', async function (req, reply) {
-    const parts = await req.multipartIterator()
```
StarpTech, line 139: If we keep `multipartIterator` at least one test should ensure the existence.

StarpTech (dismissed):

index.js
jsumners: Still a new major.

test/multipart.test.js
mcollina: done

StarpTech (approved):
LGTM

PR #151, opened 2020-08-29
Add a check if request is multipart on the preValidation hook when `attachFieldsToBody` is true.

### LINKED ISSUE #150: 406 responses for all requests with content-type other than `multipart/form-data` when attachFieldsToBody is set to true
## 🐛 Bug Report

If `attachFieldsToBody` is set to true, all requests with content-type other than `multipart/form-data` are responded with `FST_INVALID_MULTIPART_CONTENT_TYPE` error and status code 406.

## To Reproduce

```js
const fastify = require("fastify")();

fastify.register(require("fastify-multipart"), {
  attachFieldsToBody: true
});

fastify.get("/", async function (req) { 
  return req.body;
});

fastify.listen(3008, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
});

```
```bash
➜ curl http://localhost:3008
{"statusCode":406,"code":"FST_INVALID_MULTIPART_CONTENT_TYPE","error":"Not Acceptable","message":"the request is not multipart"}
```

## Expected behavior

fastify-multipart only handles requests with `multipart-form-data` content-type.

```bash
➜ curl http://localhost:3008
null
```

## Possible fix

Add a check if the current request is a multipart request before calling `req.parts()` around line 136, just like function `attachToBody`.

I can submit a PR if it is ok.

```js
if (options.attachFieldsToBody === true) {
  if (typeof options.sharedSchemaId === 'string') {
     // omit
  }
  
  fastify.addHook('preValidation', async function (req, reply) {
    if (!req.isMultipart()) { return; }  // add this
    for await (const part of req.parts()) {
      req.body = part.fields
      if (part.file) {
        if (options.onFile) {
          await options.onFile(part)
        } else {
          await part.toBuffer()
        }
      }
    }
  })
}
```

## Your Environment

- *node version*: 12
- *fastify version*: >=3.0.0
- *os*: Windows

ddadaal:
Fix #150.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,6 +133,9 @@ function fastifyMultipart (fastify, options = {}, done) {
       })
     }
     fastify.addHook('preValidation', async function (req, reply) {
+      if (!req.isMultipart()) {
+        return
+      }
       for await (const part of req.parts()) {
         req.body = part.fields
         if (part.file) {
--- a/test/multipart-attach-body.test.js
+++ b/test/multipart-attach-body.test.js
@@ -123,3 +123,38 @@ test('should be able to define a custom "onFile" handler', function (t) {
     }
   })
 })
+
+test('should not process requests with content-type other than multipart', function (t) {
+  t.plan(3)
+
+  const fastify = Fastify()
+  t.tearDown(fastify.close.bind(fastify))
+
+  fastify.register(multipart, { attachFieldsToBody: true })
+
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

StarpTech (approved):
Lgtm

PR #154, opened 2020-09-01
Update index.d.ts

BlackGlory:
Update `index.d.ts` to v3.0

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
 import * as busboy from "busboy";
 import { FastifyPlugin } from "fastify";
 import { Readable } from 'stream';
+import { FastifyError } from "fastify-error";
 
 type MultipartHandler = (
     field: string,
@@ -18,8 +19,8 @@ interface BodyEntry {
     limit: false
 }
 
-interface MultipartFields {
-  [x: string]: Multipart | Multipart[];
+export interface MultipartFields {
+    [x: string]: Multipart | Multipart[];
 }
 
 interface Multipart {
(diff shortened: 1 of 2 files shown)


StarpTech (approved):
lgtm, thanks.

mcollina (changes requested):
Would you mind to add a test for this type?

BlackGlory:
@mcollina I did not find a file for testing TypeScript types in the repo, `test/index.test-d.ts` doesn't look like a unit test.

BlackGlory:
`test/index.test-d.ts` is the v2.0 test file, so I updated it to v3.0.
But IMHO, since the file is not updated with the new version, obviously these tests are not helpful.

BlackGlory:
Considering the properties of the field type, it seems necessary to split Multipart into two interfaces.
But it will break the examples in the README, because we need to distinguish between different types. I'm not sure if these definitions should be added.

```ts
type Multipart = MultipartField | MultipartFile

interface MultipartField {
  fieldname: string
  value: string
  fieldnameTruncated: boolean
  valueTruncated: boolean
  fields: MultipartFields
}

interface MultipartFile  {
  toBuffer: () => Promise<Buffer>,
  file: NodeJS.ReadableStream,
  filepath: string,
  fieldname: string,
  filename: string,
  encoding: string,
  mimetype: string,
  fields: MultipartFields
}
```

hixus:
I might also export all interfaces (for example Multipart) so one can use them for own function declarations that use the result

mcollina:
I don't understand the question. Is this ready to land?

mcollina (approved):
lgtm

hixus:
> I don't understand the question. Is this ready to land?

Could be different pr, but for example I might wan't to define separate handler which would take just the Multipart as an argument. Currently not possible without defining it again.

```
const handleFile = async (part: Multipart) => {
  pump(part.file, fs.createWriteStream(part.filename))
}

server.post("/upload", async (req, reply) => {
  const parts = req.parts()
    if (part.file) {
      await handleFile(part)
    } else {
      console.log(part)
    }
  }
  reply.send()
});
```

BlackGlory:
This PR is ready to merge.

It may take someone more familiar with this project to decide which TypeScript interfaces should be exported, not me.

index.d.ts
```diff
@@ -18,8 +19,8 @@ interface BodyEntry {
     limit: false
 }
 
-interface MultipartFields {
-  [x: string]: Multipart;
+export interface MultipartFields {
+    [x: string]: Multipart | Multipart[];
 }
 
 interface Multipart {
```
fox1t, line 26:
```suggestion
export interface Multipart {
```

fox1t (changes requested):

fox1t:
Nice work! Thanks for you contribution. 
There is one missing export to add, then this PR can be merged.

fox1t (approved):
LGTM

mcollina (approved):
lgtm

Issue #158, opened 2020-10-01
cleanRequestFiles() trying to delete non-existent/already deleted files


bmo-at:
## 🐛 Bug Report

The cleanRequestFiles() Method tries to clean up files that have already been cleaned up previously, thereby producing a number of 'no such file or directory' errors.

## To Reproduce

Server:
```js
const server = fastify();

server.register(fastifyMultipart));

server.route({
    method: "POST",
    url: '/upload',
    handler: async (req, res) => {
        const files = await req.saveRequestFiles();
        res.send(`done. ${files.length} files received`);
    }
});
```

```sh
curl --location --request POST 'http://localhost:8080/upload' \
--form 'file=@/home/$(whoami)/testfile.txt'
```

With every request, what you'll see is fastify spitting out errors on the console proportional to the number of requests made to the server (but only if the logger is on). On the first request, no error will appear and everything will be normal. On the second request, fastify-multipart will try to clean up two files from the /tmp directory, one of which won't be found. On the third request it will be two files that can't be found, and so on. This is because the cleanRequestFiles() method will receive a growing array with a length equal to requestID, in which all but the last entry has already been deleted previously. This is beacuse the tmpUploads array which is on each request somehow also still contains the entries from previous requests. I suspect this is due to a copied object reference in decorateRequest().

## Expected behavior

I would expect every request to have its own tmpUploads array, so that separate requests can't influence each other.

## Your Environment

- *node version*: 12
- *fastify*: 3.4.1
- *fastify-multipart*: 3.2.0
- *os*: Debian 10 on WSL2

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Also cc @StarpTech for awareness.

StarpTech:
@mcollina this looks like an issue in the core. Fastify `request` must be new on every request.

mcollina:
it is new, but they all share the same array for tmpUploads. That must be reset for each request, it's a bug here.

StarpTech:
I'd strongly expect that fastify is responsible to reset the request object. This avoids such memory leaks.

StarpTech:
But yes, in order to fix it we need to reset the array in the `cleanRequestFiles` function.

PR #159, opened 2020-10-01
Reset tmpUploads at every request

mcollina:
Fixes https://github.com/fastify/fastify-multipart/issues/158

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -174,7 +174,7 @@ function fastifyMultipart (fastify, options = {}, done) {
   fastify.decorateRequest('multipartIterator', getMultipartIterator)
 
   fastify.decorateRequest('isMultipart', isMultipart)
-  fastify.decorateRequest('tmpUploads', [])
+  fastify.decorateRequest('tmpUploads', null)
 
   // legacy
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -491,6 +492,9 @@ function fastifyMultipart (fastify, options = {}, done) {
   }
 
   async function cleanRequestFiles () {
+    if (!this.tmpUploads) {
```
zekth, line 495: Particular reason to not set this condition in the `onResponse` hook?

mcollina: I check if this exists before using it.

zekth: if mean instead of doing this:
```javascript
  fastify.addHook('onResponse', async (request, reply) => {
    if (request.tmpUploads) {
      await request.cleanRequestFiles()
    }
  })
```

mcollina: It's not possible to do so as `cleanRequestFIles` is publicly exposed, and I think it's best to protect it.

StarpTech:
I think this is ready to merge.

StarpTech:
Your t-shirt is on the way! 😄

PR #161, opened 2020-10-03
rethrow busboy error when storing files on disk, add tests

### LINKED ISSUE #160: Inconsistent behavior when handling file limits
## 🐛 Bug Report

The limits you can give to fastify-multipart are handeled in an inconsistent way.

If, for example, you give it a limit for the number of files it will automatically end the request with a 413 or alternatively I can catch the error that saveRequestFiles() throws and create my own response on top of that.

However, if I give it a limit for the file size it will not do the same.

In the docs it says to catch the errors associated with file limits. But a file that is too large never throws an error up to the application level, instead it is logged twice, once as a warning and then again as an error down in the saveRequestFiles() method.

It never bubbles up and is not catchable for me.

The problem here is that I can then never properly end the request and the client is just stuck.

## To Reproduce

```js
const server = require('fastify')({ logger: true });

server.register(require('fastify-multipart'), { limits: { files: 1, fileSize: 1 * 1000 * 1000 } });

server.route({
    method: "POST",
    url: '/upload',
    handler: async (req, res) => {

        let files ;

        try {
            files = await req.saveRequestFiles();
        } catch (error) {
            req.log.error(error);
            res.status(500).send('error');
        }
        if (!res.sent) { res.send(`done. ${files.length} files received`); }
    }
});

server.listen(8080, (error) => {
    if (error) {
        server.log.error(error);
        process.exit(1);
    }
})
```

## Expected behavior

This server should accept requests of up to one file per request with a maximum size of 1 megabyte. 

saveRequestFiles() should throw an error whenever it runs into a file limit which can then be handled by the developer.

Alternatively, if the developer does not handle the error, it should end the request with a 413 code like it already does for the limit for number of files.

## Your Environment

- *node version*: 12
- *fastify:* 3.4.1
- *fastify-multipart*: 3.2.0
- *os*: Debian 10 on WSL2

StarpTech:
Fix #160 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ If you do set upload limits, be sure to catch the error. An error or exception w
 **Note**: if the file stream that is provided by `data.file` is not consumed (like in the example below with the usage of pump) the promise won't be fulfilled at the end of the multipart processing.
 This behavior is inherited from [busboy](https://github.com/mscdex/busboy).
 
-**Note**: if you set a `fileSize` limit and you want to know if the file limit was reached you can listen to `data.file.on('limit')` or check at the of the stream the property `data.file.truncated`. 
+**Note**: if you set a `fileSize` limit and you want to know if the file limit was reached you can listen to `data.file.on('limit')` or check at the end of the stream the property `data.file.truncated`. 
 
 ```js
 try {
--- a/README.md
+++ b/README.md
@@ -160,8 +160,6 @@ fastify.post('/upload/raw/any', async function (req, reply) {
 
 This will store all files in the operating system default directory for temporary files. As soon as the response ends all files are removed.
 
-**Note**: if you set a `fileSize` limit `req.saveRequestFiles()` is able to throw an `RequestFileTooLargeError` error.
-
 ```js
 fastify.post('/upload/files', async function (req, reply) {
   // stores files to tmp dir and return files
@@ -174,6 +172,21 @@ fastify.post('/upload/files', async function (req, reply) {
   files[0].fields // other parsed parts
 
   reply.send()
+})
+```
+
+## Handle file size limitation
+
+If you set a `fileSize` limit `req.saveRequestFiles()` is able to throw an `RequestFileTooLargeError` error.
+
+```js
+fastify.post('/upload/files', async function (req, reply) {
+  try {
+    const files = await req.saveRequestFiles({ limits: { fileSize: 17000 } })
+    reply.send()
+  } catch (error) {
+    // error instanceof fastify.multipartErrors.RequestFileTooLargeError
+  }
 })
 ```
 
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ If you do set upload limits, be sure to catch the error. An error or exception w
 **Note**: if the file stream that is provided by `data.file` is not consumed (like in the example below with the usage of pump) the promise won't be fulfilled at the end of the multipart processing.
 This behavior is inherited from [busboy](https://github.com/mscdex/busboy).
 
+**Note**: if you set a `fileSize` limit and you want to know if the file limit was reached you can listen to `data.file.on('limit')` or check at the of the stream the property `data.file.truncated`. 
+
 ```js
 try {
   const data = await req.file()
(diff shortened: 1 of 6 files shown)


StarpTech:
Lovely works on my machine but fails here.

test/multipart-disk.test.js
```diff
@@ -78,6 +80,135 @@ test('should store file on disk, remove on response', function (t) {
...
+      await req.saveRequestFiles({ limits: { fileSize: 500 } })
+      reply.code(200).send()
+    } catch (error) {
+      t.true(error instanceof fastify.multipartErrors.RequestFileTooLargeError)
+      t.equal(error.part.fieldname, 'upload')
+      reply.code(500).send()
+    }
+  })
+
+  fastify.listen(0, async function () {
```
mcollina, line 104: passing an async function here would swallow the error. Use async/await (see my other PR for a fix).

StarpTech: Why is it ok to swallow the error?

mcollina: it is not ok. The error is being swallowed because you pass an async function in a place that's not meant to receive an async function.

mcollina:
See https://github.com/fastify/fastify-multipart/pull/159 for how to fix the callback problem.

StarpTech:
I'm not sure why this should fix the test. In callback mode tap will wait until the plan count is reached.

Anyway, it doesn't help.

mcollina:
> I'm not sure why this should fix the test. In callback mode tap will wait until the plan count is reached.

It does not fix the test, but it let some errors bubble up that otherwise will be swallowed.

StarpTech:
> It does not fix the test, but it let some errors bubble up that otherwise will be swallowed.

Ah, ok. Any idea of how to fix the test? Feels like a race condition. My PC is too fast to reproduce.

mcollina:
I enabled some logging. It seems only Node v10 has this problem. I'm not really sure what is the problem, but we are likely missing some error event handler somewhere.

StarpTech:
> I enabled some logging. It seems only Node v10 has this problem. I'm not really sure what is the problem, but we are likely missing some error event handler somewhere.

I don't think so. It failed previously on 10 and 14.

StarpTech:
@mcollina I can reproduce by testing only these tests in one file. When I split them in two it works. I look like that one instance influence the other. Ideas?

```
'should store file on disk, remove on response error'
'should throw on file limit error'
```

StarpTech:
Perhaps related https://github.com/atian25/egg-showcase/pull/18 it failed with the same connection error. Busboy has no ci for windows or mac.

StarpTech:
Not sure if it is the right decision to rely on busboy. It looks unmaintained.

mcollina:
I think we should always add a `on('error')` handler here.

StarpTech:
How do you want to propagate the error then?

StarpTech:
@mcollina ping

mcollina:
What I said in https://github.com/fastify/fastify-multipart/pull/161#issuecomment-704281873, is that you are likely getting that error because you are missing an error handler in that block. An error can happen for all sort of reasons, and no one is handling it.

StarpTech:
I tried all your suggestions. I have no ideas anymore. I will keep this open PR.

StarpTech:
@salmanm could you reference the build that run successfully?

StarpTech:
According to the busboy documentation https://github.com/mscdex/busboy/blob/baae157b97e4c4bc942a453e51168b2ae258879d/README.md#busboy-special-events `limit` should be checked at the end of the stream. We tried to do it in between and it resulted in an unmanageable condition. The current implementation make some different assumptions.

- File limit event must be handled manually if the user works with streams.
- File limit error is handled automatically when working in disk mode.

@mcollina

index.js
```diff
@@ -482,9 +449,20 @@ function fastifyMultipart (fastify, options = {}, done) {
...
+          err.part = file
+          throw err
+        }
       } catch (error) {
-        this.log.error(error)
-        await unlink(filepath)
+        try {
+          await unlink(filepath)
+        } catch (error) {
+          this.log.debug('saveRequestFiles: could not delete file, %s', error.messsage)
```
mcollina, line 462: I would log this as `this.log.debug({ err }, 'could not delete file')`.

README.md
```diff
@@ -158,6 +160,8 @@ fastify.post('/upload/raw/any', async function (req, reply) {
 
 This will store all files in the operating system default directory for temporary files. As soon as the response ends all files are removed.
 
+**Note**: if you set a `fileSize` limit `req.saveRequestFiles()` is able to throw an `RequestFileTooLargeError` error.
```
mcollina, line 163: Can you add an example on how to handle this?

mcollina (commented):
+1 with the approach, just a couple of nits

mcollina (approved):
lgtm

PR #163, opened 2020-10-12
fix issue that in multipart handler might miss fields

JacobChang:
This is to fix one issue that I found in my project, not sure whether this behaviour is intended or not.

Basically, I found that in some cases, it's possible that onField of busyboy will be trigger multiple times before user read the parsed lastValue, this will cause some field in formdata are missing.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -292,26 +292,33 @@ function fastifyMultipart (fastify, options = {}, done) {
 
     this.log.debug('starting multipart parsing')
 
-    let worker
-    let lastValue
+    let values = []
+    let pendingHandler = null
 
     // only one file / field can be processed at a time
     // "null" will close the consumer side
     const ch = (val) => {
-      if (typeof val === 'function') {
-        worker = val
+      if (pendingHandler) {
+        pendingHandler(val)
+        pendingHandler = null
       } else {
-        lastValue = val
-      }
-      if (worker && lastValue !== undefined) {
-        worker(lastValue)
-        worker = undefined
-        lastValue = undefined
+        values.push(val)
       }
     }
+
+    const handle = (handler) => {
+      if (values.length > 0) {
+        const value = values[0]
+        values = values.slice(1)
+        handler(value)
+      } else {
+        pendingHandler = handler
+      }
+    }
+
     const parts = () => {
       return new Promise((resolve, reject) => {
-        ch((val) => {
+        handle((val) => {
           if (val instanceof Error) return reject(val)
           resolve(val)
         })
--- a/test/multipart.test.js
+++ b/test/multipart.test.js
@@ -544,6 +544,7 @@ test('should receive all fields', async function (t) {
   t.plan(11)
 
   const original = fs.readFileSync(filePath, 'utf8')
+  const immediate = util.promisify(setImmediate)
 
   const fastify = Fastify()
   t.tearDown(fastify.close.bind(fastify))
@@ -571,7 +572,7 @@ test('should receive all fields', async function (t) {
             t.equal(buf.toString(), original)
           })
         )
-        await new Promise(setImmediate)
+        await immediate()
       }
 
       recvField[part.fieldname] = true
--- a/test/multipart.test.js
+++ b/test/multipart.test.js
@@ -10,6 +10,7 @@ const path = require('path')
 const fs = require('fs')
 const concat = require('concat-stream')
 const stream = require('stream')
+const { once } = require('events')
 const pump = util.promisify(stream.pipeline)
 const sendToWormhole = require('stream-wormhole')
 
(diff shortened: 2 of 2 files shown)


mcollina (changes requested):
Can you please add a unit test?

JacobChang:
@mcollina done, add a simple unit test to reproduce this issue

test/multipart.test.js
```diff
@@ -538,3 +538,81 @@ test('should also work with multipartIterator', function (t) {
...
+        const original = fs.readFileSync(filePath, 'utf8')
+        await pump(
+          part.file,
+          concat(function (buf) {
+            t.equal(buf.toString(), original)
+          })
+        )
+        // wait for 1s to mimic situation that user side takes a long time to process every part
+        await new Promise((resolve, reject) => {
+          setTimeout(resolve, 1000)
```
mcollina, line 575: Please avoid using timers in tests.

test/multipart.test.js
```diff
@@ -538,3 +538,81 @@ test('should also work with multipartIterator', function (t) {
...
+    }
+
+    t.equal(recvField.upload, true)
+    t.equal(recvField.hello, true)
+    t.equal(recvField.willbe, true)
+
+    reply.code(200).send()
+  })
+
+  fastify.listen(0, async function () {
```
mcollina, line 587: Please do not pass async here and use the promise form of fastify.listen()

mcollina (changes requested):

test/multipart.test.js
JacobChang: Then what's the standard way to simulate a time-consuming async action ?

test/multipart.test.js
JacobChang: Okay, I copy from other tests, seems all tests use async in this place

mcollina: Which is incorrect anyway, just trying to avoid adding even more problems.

test/multipart.test.js
```diff
@@ -538,3 +538,81 @@ test('should also work with multipartIterator', function (t) {
...
+    for await (const part of req.parts()) {
+      recvField[part.fieldname] = true
+      if (part.file) {
+        t.equal(part.fieldname, 'upload')
+        t.equal(part.filename, 'README.md')
+        t.equal(part.encoding, '7bit')
+        t.equal(part.mimetype, 'text/markdown')
+        t.ok(part.fields.upload)
+
+        const original = fs.readFileSync(filePath, 'utf8')
```
mcollina, line 566: could you move this at the top of the test?

test/multipart.test.js
mcollina: It's dependent on each case. I think just using a setImmediate would work in this case.

test/multipart.test.js
mcollina: See https://github.com/fastify/fastify-multipart/pull/164 for an example.

JacobChang: Thx, updated

test/multipart.test.js
```diff
@@ -555,26 +558,23 @@ test('should receive all field', function (t) {
...
           part.file,
           concat(function (buf) {
             t.equal(buf.toString(), original)
           })
         )
-        // wait for 1s to mimic situation that user side takes a long time to process every part
-        await new Promise((resolve, reject) => {
-          setTimeout(resolve, 1000)
-        })
+        await new Promise(setImmediate)
```
mcollina, line 574: please use `const immediate = promisify(setImmediate)` at the top.

mcollina (approved):
lgtm

StarpTech (changes requested):

index.js
```diff
@@ -292,26 +292,33 @@ function fastifyMultipart (fastify, options = {}, done) {
...
-      if (worker && lastValue !== undefined) {
-        worker(lastValue)
-        worker = undefined
-        lastValue = undefined
+    }
+
+    const handle = (handler) => {
+      if (values.length > 0) {
+        const value = values[0]
+        values = values.slice(1)
```
StarpTech, line 312: Why not use `handler(values.shift())` ?

StarpTech (changes requested):

index.js
JacobChang: Just want to make the logic more explicit. The same reason that I use function *cb* for values and *handle* for callback function instead of using one function *cb* to handle both values and callback function.

test/multipart.test.js
```diff
@@ -538,3 +539,79 @@ test('should also work with multipartIterator', function (t) {
     }
   })
 })
+
+test('should receive all fields', async function (t) {
```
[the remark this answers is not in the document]
JacobChang: updated, thx for the review

StarpTech (approved):
LGTM

PR #165, opened 2020-10-13
Add correct type for non-file value

### LINKED ISSUE #162: Inconsistent Multipart type when some fields are not files
# 🐛 Bug Report

When sending multiple fields in the form-data, Typescript types them all as files, even if it's not the case

## To Reproduce

```js
server.post<{Body: {field1: Multipart, field2:Multipart}}>(...)
```

```
http -f :3000/mulitpart field1=foo field2@myFile.txt
```
With this call, we submit a file and a string. But due to the [Multipart type](https://github.com/fastify/fastify-multipart/blob/master/index.d.ts#L26), they're all typed the same and we can't access the `.value` property containing the string (as in this [test](https://github.com/fastify/fastify-multipart/blob/master/test/multipart-body-schema.test.js#L45)).

## Expected behavior

We could take advantage of TypeScript [conditional typing](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) to allow user to properly type its body.

```typescript
interface MultipartFile {
    filename: string;
}
interface MultipartString {
    value: string;
}
type Multipart<T = true> = T extends string ? MultipartString : MultipartFile;

const mult = {} as Multipart<string>
mult.value // OK
mult.filename // error

const multFile = {} as Multipart
multFile.value // error
multFile.filename // OK
```
[Playground to see it in action](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLIFcA2ZgAc5RgBiwmKA3gLABQydyMpEIcAthAFzIDOYUoAcwDcNAL41QkWIhQZseAmADKfQciq16ANziZ0nHqpDCxNMAE9csrDnyEAPABVkAXmR99APlfJnEAB6QIAAm3Ib8xsgA-Gg2CoQqEQLIXHK2iiRkItQ0CAD2ILzIrDY+5KLIcGFp8WD2vEmeNCXYAHQ6eigA9F3IAPIA0s02rYxkLOzIPcjQUHlQNLkFRS3ETGUVVbHydmDD2JkQ7br6U72z8-trZKNME929g0A)

## Your Environment

- *node version*: 14
- *TypeScript version*: 4.0.2
- *fastify version*: 3.3.0
- *fastiify-multipart version*: 3.2.0
- *os*: Linux

JbIPS:
Allow user to declare a field as a non-file value with a generic
`Multipart<T>` to set the correct type of the value.

Closes #162

### DIFF
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -1,5 +1,6 @@
 import fastify from 'fastify'
-import fastifyMultipart, {Multipart, MultipartFields} from '..'
+import fastifyMultipart from '..'
+import { Multipart, MultipartFields } from '..'
 import * as util from 'util'
 import { pipeline } from 'stream'
 import * as fs from 'fs'
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -1,10 +1,9 @@
 import fastify from 'fastify'
-import fastifyMultipart from '..'
-import { MultipartFields } from '..'
+import fastifyMultipart, {Multipart, MultipartFields} from '..'
 import * as util from 'util'
 import { pipeline } from 'stream'
 import * as fs from 'fs'
-import { expectType } from 'tsd'
+import { expectError, expectType } from 'tsd'
 
 const pump = util.promisify(pipeline)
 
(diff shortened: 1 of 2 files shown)


mcollina:
Can you please add a unit test?

Ethan-Arrowood (changes requested):
please add a test using TSD - otherwise this looks good thank you!

JbIPS:
First time with TSD so tell me if it's not enough

test/index.test-d.ts
```diff
@@ -1,10 +1,9 @@
 import fastify from 'fastify'
-import fastifyMultipart from '..'
-import { MultipartFields } from '..'
+import fastifyMultipart, {Multipart, MultipartFields} from '..'
```
fox1t, line 2: Seems that formatting is missing here

fox1t (commented):
LGTM

test/index.test-d.ts
JbIPS: Sorry, fixed that!

mcollina (approved):
lgtm

mcollina:
@Ethan-Arrowood you good to land this?

Ethan-Arrowood (approved):
LGTM!

PR #169, opened 2020-10-19
add install instruction for ts types

### LINKED ISSUE #168: Plugin throws an error when transpiling using typescript
## 🐛 Bug Report

When using this plugin with typescript it throws an error when transpiling my app using `tsc`. The error says
```
Could not find a declaration file for module 'busboy'. 
'[..]/node_modules/busboy/lib/main.js' implicitly has an 'any' type.
Try `npm install @types/busboy` if it exists or add a new declaration (.d.ts) file containing `declare module 'busboy';`
1 import * as busboy from "busboy";
```

## To Reproduce

```ts
// index.ts
import Fastify from 'fastify'
import multiPart from 'fastify-multipart'

const fastify = Fastify({
  logger: true,
})

fastify.register(multiPart)

fastify.get('/', async (request, reply) => {
  return { hello: 'world' }
})

fastify.listen(3000, (err, address) => {
  if (err) throw err
  fastify.log.info(`server listening on ${address}`)
})

```
then run `tsc`

## Expected behavior
The plugin should either ship the correct typings for busboy as well or should say to install them as a dev dependency in the README

## Your Environment

- *node version*: v12.19.0
- *fastify version*: 3.7.0
- *os*: Linux

StarpTech:
Should fix #168 

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,11 @@
     "fastify-error": "^0.2.0",
     "fastify-plugin": "^3.0.0",
     "hexoid": "^1.0.0",
-    "stream-wormhole": "^1.1.0",
-    "@types/busboy": "^0.2.3"
+    "stream-wormhole": "^1.1.0"
   },
   "devDependencies": {
     "@types/node": "^14.0.27",
+    "@types/busboy": "^0.2.3",
     "@typescript-eslint/parser": "^4.0.0",
     "climem": "^1.0.3",
     "eslint": "^7.7.0",
--- a/package.json
+++ b/package.json
@@ -11,10 +11,10 @@
     "fastify-error": "^0.2.0",
     "fastify-plugin": "^3.0.0",
     "hexoid": "^1.0.0",
-    "stream-wormhole": "^1.1.0"
+    "stream-wormhole": "^1.1.0",
+    "@types/busboy": "^0.2.3"
   },
   "devDependencies": {
-    "@types/busboy": "^0.2.3",
     "@types/node": "^14.0.27",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,10 +11,10 @@
     "fastify-error": "^0.2.0",
     "fastify-plugin": "^3.0.0",
     "hexoid": "^1.0.0",
-    "stream-wormhole": "^1.1.0"
+    "stream-wormhole": "^1.1.0",
+    "@types/busboy": "^0.2.3"
```
mcollina, line 15: I would prefer not shipping anything from definitelyTyped as they are often incomplete and out of the control of the maintainer. I would just let folks know to install them as devDeps when using ts.

mcollina (changes requested):

package.json
Ethan-Arrowood: As we have in other plugins ( as well as on fastify core ) we do not add typescript types to the dependency list. The recommended solution is to add these types to the readme and instruct users to install them (sorta like a peer dep but we also don't use that npm feature because of its inconsistent behavior).

I know this isn't the best solution but its what we compromised and we should continue to stick to it until something better comes along in the future 😁

StarpTech: https://github.com/fastify/fastify/blob/0501b3a189b309fc6e81b00046f80efdccdfd521/types/logger.d.ts#L2-L11

Then I prefer to duplicate the type definition as in fastify core.

mcollina (approved):
lgtm

PR #177, opened 2020-10-29
saveRequestFiles() now writes non-file fields to body instead of discarding them

bmo-at:
saveRequestFiles() used to just discard andy non-file parts of the body.
This was due to it only using the files iterator on the request.
I have changed it to use the parts iterator instead. The behavior for files is unchanged, but non-file parts now get written to the body. This was curiously undefined in all the scenarios I encountered, so I also added a check for undefined to make the body an empty object in case it is undefined.

This pull request should fix the behavior I asked a question about in https://github.com/fastify/help/issues/286

cc @mcollina, took me a little bit but here it is

I did my best but it still feels a little weird, so I'd be thankful for criticism
 

#### Checklist

Tests: 

 PASS  test/multipart-attach-body.test.js 16 OK 413.857ms
 PASS  test/multipart-big-stream.test.js 3 OK 527.862ms
 PASS  test/multipart-concat.test.js 9 OK 264.458ms
 PASS  test/legacy/multipart.test.js 43 OK 365.88ms
 PASS  test/multipart-body-schema.test.js 8 OK 415.456ms
 PASS  test/legacy/append-body.test.js 74 OK 1s
 PASS  test/multipart-security.test.js 8 OK 277.759ms
 PASS  test/multipart-disk.test.js 42 OK 498.296ms
 PASS  test/multipart-small-stream.test.js 2 OK 168.481ms
 PASS  test/multipart.test.js 68 OK 547.342ms
 PASS  test/big.test.js 9 OK 10s
 PASS  test/legacy/big.test.js 12 OK 14s

  🌈 SUMMARY RESULTS 🌈

Suites:   12 passed, 12 of 12 completed
Asserts:  294 passed, of 294
Time:     19s
----------|----------|----------|----------|----------|-------------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------|----------|----------|----------|----------|-------------------|
All files |    95.82 |    85.85 |      100 |    95.74 |                   |
 index.js |    95.82 |    85.85 |      100 |    95.74 |... 74,475,477,493 |
----------|----------|----------|----------|----------|-------------------|
[master 83dfd1b] use parts iterator instead of files and write non-file parts onto the body in saveRequestFiles()
 1 file changed, 26 insertions(+), 15 deletions(-)

### DIFF
--- a/index.js
+++ b/index.js
@@ -447,24 +447,35 @@ function fastifyMultipart (fastify, options = {}, done) {
   async function saveRequestFiles (options) {
     const requestFiles = []
 
-    const files = await this.files(options)
+    const parts = await this.parts(options)
     this.tmpUploads = []
-    for await (const file of files) {
-      const filepath = path.join(os.tmpdir(), toID() + path.extname(file.filename))
-      const target = createWriteStream(filepath)
-      try {
-        await pump(file.file, target)
-        requestFiles.push({ ...file, filepath })
-        this.tmpUploads.push(filepath)
-        // busboy set truncated to true when the configured file size limit was reached
-        if (file.file.truncated) {
-          const err = new RequestFileTooLargeError()
-          err.part = file
+    if (this.body === undefined) { this.body = {} }
+    for await (const part of parts) {
+      if (part.file) {
+        const filepath = path.join(os.tmpdir(), toID() + path.extname(part.filename))
+        const target = createWriteStream(filepath)
+        try {
+          await pump(part.file, target)
+          requestFiles.push({ ...part, filepath })
+          this.tmpUploads.push(filepath)
+          // busboy set truncated to true when the configured file size limit was reached
+          if (part.file.truncated) {
+            const err = new RequestFileTooLargeError()
+            err.part = part
+            throw err
+          }
+        } catch (err) {
+          this.log.error({ err }, 'save request file')
           throw err
         }
-      } catch (err) {
-        this.log.error({ err }, 'save request file')
-        throw err
+      } else {
+        if (this.body[part.fieldname] === undefined) {
+          this.body[part.fieldname] = part.value
+        } else if (Array.isArray(this.body[part.fieldname])) {
+          this.body[part.fieldname].push(part.value)
+        } else {
+          this.body[part.fieldname] = [this.body[part.fieldname], part.value]
+        }
       }
     }
 


StarpTech:
Hi @bmo-at thanks for the PR. You can access other fields in this way: https://github.com/fastify/fastify-multipart#upload-files-to-disk-and-work-with-temporary-file-paths

index.js
```diff
@@ -447,24 +447,35 @@ function fastifyMultipart (fastify, options = {}, done) {
...
-          const err = new RequestFileTooLargeError()
-          err.part = file
+    if (this.body === undefined) { this.body = {} }
+    for await (const part of parts) {
+      if (part.file) {
+        const filepath = path.join(os.tmpdir(), toID() + path.extname(part.filename))
+        const target = createWriteStream(filepath)
+        try {
+          await pump(part.file, target)
+          requestFiles.push({ ...part, filepath })
```
StarpTech, line 459: Your result can be archived by a few lines:

```js
if (this.body === undefined) {
    this.body = part.fields
} else {
   this.body = { ...this.body, ...part.fields } 
}
```

StarpTech (changes requested):

index.js
bmo-at: I agree you can achieve a similar result, but with that you also write back all the files to the body, as they are part of the part.fields attribute.
IMHO this shouldn't be done; if you use saveRequestFiles to write the files to disk they shouldn't also be added to the body. Only the non-file attributes should remain in the body, as they need to be accessible.

StarpTech: I agree, then only add non-file attributes.

```js
if (this.body === undefined) {
    this.body = part.fields.map(x => !x.file)
} else {
   this.body = { ...this.body, ...part.fields.map(x => !x.file) } 
}
```

bmo-at: Nice use of map. This looks extremely clean. I'll fix it up this way and submit it.

bmo-at: @StarpTech Ooof well this was a wild ride. part.fields is an object, not an array.
Honestly I'd recommend sticking with my original code. Everything else is just shitshow of case handling.

StarpTech: What's wrong with that?
```js
for (var fieldName in part.fields) {
  const field = this.body[fieldName]
  if (this.body[fieldName].hasOwnProperty(fieldName) && !field.file) {
	this.body[fieldName] = field
  }
}
```

bmo-at: What irks me about this is the fact the we are already iterating over all the parts and accessing the .fields property. This only needs to happen once, but this way it happens on every file in the request.
Imagine a scenario where 10 files are in a request, then this is executed ten times where one would have sufficed...

StarpTech: I agree, I'm also fine with.

```js
this.body = { ...this.body, ...part.fields }
```

StarpTech (changes requested):

index.js
StarpTech: Then just do:

```js
this.body = { ...this.body, ...part.fields } 
```

bmo-at: So then we are back at square zero where we have a duplication of information: On the one hand, the files from the multipart form have been written to disk, on the other hand they are accessible through the body again.

What exactly is so bad about using the parts iterator and my approach in adding the fields to the body? To me it seems like the only logical thing given the desired result of having only the non-file fields in the body and doing so without a performance penalty for requests with a large number of parts.

StarpTech: It looks odd because we already collected all fields before in the same way. https://github.com/fastify/fastify-multipart/blob/ccf96b0cd138593fd4b1ce1c0bcf2a46ac816cb2/index.js#L423-L429

PR #191, opened 2020-12-04
Move @types/busboy to dependencies

erfanium:
I can't use this package in Typescript, because every time I try to build my project via `tsc`, compiler shows an error that package 'busboy' in `index.d.ts` was not found.

I'm not sure if this change is correct or not, But I think it's better to consider `index.d.ts` types dependencies as dependency (not dev-dependency)

Thanks!

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
+    "@types/busboy": "^0.2.3",
     "busboy": "^0.3.1",
     "deepmerge": "^4.2.2",
     "end-of-stream": "^1.4.4",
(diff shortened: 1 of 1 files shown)


kibertoad:
Just add types for busboy as devDependencies in your project. Documentation update for that would be much appreciated. We don't bundle types as runtime dependencies in our libs.

erfanium:
@kibertoad Sure!, I'll update the doc

> We don't bundle types as runtime dependencies in our libs.

I was curious, what is the reason for this?

mcollina:
> I was curious, what is the reason for this?

A lot of folks complain about install sizes and the amount of dependencies. Given that types are development-only, those are not really needed for production usage.

PR #198, opened 2021-02-01
fix: file size limit

### LINKED ISSUE #196: fileSize limit on fastify-multipart did not throw error
I don't know if I use it correctly or not but it is just pass through if I passing file with size higher than limit.

```js
/*
 * other code
 */

fastify.post('/upload', async function (req, reply) {
    const data = await req.file({
      limits: {
        fileSize: 1 * 1024 * 1024,  // 1mb limit
      }
    })

    await pump(data.file, fs.createWriteStream(data.filename))

    reply.send()
})
```

It write file to my directory, when I expect it to throw error

* *node version*: 14.15.3
* *fastify version*: 3.9.2
* *fastify multipart version*: 3.3.1
* *os*: WSL

climba03003:
This PR resolve #196 by adding a new option `throwFileSizeLimit`.

The option can be configure in two way:

#### Globally
```JS
fastify.register(fastifyMultipart, { throwFileSizeLimit: false })
```

#### Per Functions
```JS
const file = await req.file({ throwFileSizeLimit: false, limits: { fileSize: 17000 } })
const files = await req.files({ throwFileSizeLimit: false, limits: { fileSize: 17000 } })
const parts = await req.parts({ throwFileSizeLimit: false, limits: { fileSize: 17000 } })
```

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -174,9 +174,6 @@ function fastifyMultipart (fastify, options, done) {
   fastify.decorateRequest(kMultipartHandler, handleMultipart)
 
   fastify.decorateRequest('parts', getMultipartIterator)
-  // keeping multipartIterator to avoid bumping a major
-  // TODO remove on 4.x
-  fastify.decorateRequest('multipartIterator', getMultipartIterator)
 
   fastify.decorateRequest('isMultipart', isMultipart)
   fastify.decorateRequest('tmpUploads', null)
--- a/index.js
+++ b/index.js
@@ -149,6 +149,11 @@ function fastifyMultipart (fastify, options, done) {
     })
   }
 
+  let throwFileSizeLimit = true
+  if (typeof options.throwFileSizeLimit === 'boolean') {
+    throwFileSizeLimit = options.throwFileSizeLimit
+  }
+
   const PartsLimitError = createError('FST_PARTS_LIMIT', 'reach parts limit', 413)
   const FilesLimitError = createError('FST_FILES_LIMIT', 'reach files limit', 413)
   const FieldsLimitError = createError('FST_FIELDS_LIMIT', 'reach fields limit', 413)
@@ -400,7 +405,11 @@ function fastifyMultipart (fastify, options, done) {
         return
       }
 
-      if (opts.throwFileSizeLimit || options.throwFileSizeLimit) {
+      if (typeof opts.throwFileSizeLimit === 'boolean') {
+        throwFileSizeLimit = opts.throwFileSizeLimit
+      }
+
+      if (throwFileSizeLimit) {
         file.on('limit', function () {
           onError(new RequestFileTooLargeError())
         })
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -400,6 +400,12 @@ function fastifyMultipart (fastify, options, done) {
         return
       }
 
+      if (opts.throwFileSizeLimit || options.throwFileSizeLimit) {
```
mcollina, line 403: I would consider those default to `true`. This seems a bug in the original implementation.

mcollina (changes requested):

index.js
climba03003: We can release a `semver-minor` with false, then go with a `semver-major` with true in another PR.
So, it would not break someone code and allow them to use this feature.

zekth: agree with @climba03003 , this bug was there for a certain amount of time so maybe people would break their code if the upgrade is in `semver-minor`

zekth: side node `opts` and `options` are kind of confusing. Might be worth to rename those properly.

climba03003: I agree that `opts` and `options` are too confusing when I scanning the code. I will take a time to improve this same as what I have done for fastify-swagger. I will mark it as `TODO`. 

The big `refactor` will comes with `semver-major` and the default behavior change.

mcollina: I'm happy to bump the major and have a good default.

mcollina (approved):
lgtm

StarpTech:
With that change we can remove the check here https://github.com/fastify/fastify-multipart/blob/5529808b5a97626c9661ecc8683cc1f8a8411e5c/index.js#L465-L469

StarpTech:
@mcollina I think it's the right moment to remove the old callback-API when we bump to a major.

mcollina:
While it pollutes the code, I think it can stay here for a bit longer - there are plenty of people using it.

climba03003:
> 
> 
> With that change we can remove the check here
> 
> https://github.com/fastify/fastify-multipart/blob/5529808b5a97626c9661ecc8683cc1f8a8411e5c/index.js#L465-L469

Do you have any clue why this two cases will fail when I deleted the code you mention?
https://github.com/fastify/fastify-multipart/blob/master/test/multipart-disk.test.js#L140-L261

StarpTech:
> While it pollutes the code, I think it can stay here for a bit longer - there are plenty of people using it.

Version will still exist in the registry.

mcollina:
> Version will still exist in the registry.

The fact that an API is not maintained anymore will cause significant friction to some of our old user base. Given that it has not cause many issues so far, I think it can stay there as it is.

StarpTech:
> The fact that an API is not maintained anymore will cause significant friction to some of our old user base. Given that it has not cause many issues so far, I think it can stay there as it is.

Reasonable but hard to measure. I'd prefer a "legacy" branch to split those interest groups.

mcollina:
@StarpTech anything blocking here?

StarpTech (approved):
LGTM

delvedor (approved):
LGTM

PR #213, opened 2021-03-12
Test calling finished if one of the streams closes prematurely

Vunovati:
Cover the case when one of the streams gets closed prematurely after Fastify had received it.

#### Checklist

### DIFF
--- a/test/legacy/multipart.test.js
+++ b/test/legacy/multipart.test.js
@@ -139,6 +139,78 @@ test('should call finished when both files are pumped', { skip: process.platform
   })
 })
 
+test('should call finished if one of the streams closes prematurely', { skip: process.platform === 'win32' }, function (t) {
+  t.plan(5)
+
+  const fastify = Fastify({
+    logger: {
+      level: 'debug'
+    }
+  })
+  t.tearDown(fastify.close.bind(fastify))
+
+  fastify.register(multipart)
+
+  fastify.post('/', function (req, reply) {
+    let fileCount = 0
+    t.ok(req.isMultipart())
+
+    req.multipart(handler, function () {
+      t.equal(fileCount, 1)
+      reply.code(200).send()
+    })
+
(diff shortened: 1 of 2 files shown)


Vunovati:
I was still unable to cover the last remaining two lines (282,283) with tests. Here are my premises and conclusion why I think that is.

Two lines in `waitForFiles` are executed when:
`!err && !completed && callDoneOnNextEos && (count === files)`

According to my understanding (and observations), busboy `finish` event will only be fired after the file streams it contains have been closed. EOS’s callback, `waitForFiles`, will be invoked after those file streams have ended. 
That means that busboy finish event will always be fired after all the `waitForFiles` has been called.

Since busboy finish callback is the only one that can set the `callDoneOnNextEos` to true, the aforementioned order of events would mean that `callDoneOnNextEos` will never be true when evaluated in waitForFiles and testing lines 282,283 would not be possible:
```
      if (callDoneOnNextEos && count === files) {
        completed = true
        done()
      }
```
If I am correct, then these lines would be unreachable for all cases.

mcollina:
I understand! Go ahead and simplify all that logic removing both `count` and `callDoneOnNextEos`.

test/legacy/multipart.test.js
```diff
@@ -139,6 +139,78 @@ test('should call finished when both files are pumped', { skip: process.platform
   })
 })
 
+test('should call finished if one of the streams closes prematurely', { skip: process.platform === 'win32' }, function (t) {
+  t.plan(5)
+
+  const fastify = Fastify({
+    logger: {
+      level: 'debug'
+    }
```
simoneb, line 148: remove logging as we don't want the tests to output anything

simoneb (approved):
One comment, LGTM

test/legacy/multipart.test.js
Vunovati: Done

simoneb (approved):

PR #218, opened 2021-04-05
chore tap 15

Eomm:
This is an automatic PR created with [massive-wax](https://github.com/Eomm/massive-wax)!


mcollina:
CI fails

zekth:
FIxed the flag but cannot reproduce the test failing on my current machine.

mcollina:
You need to set `check-coverage: false` in `.taprc`

Eomm:
There is a test skipped on windows that I can't get pass :/

mcollina:
> There is a test skipped on windows that I can't get pass :/

I'm lost. Is this failing elsewhere? Are you trying to get this test to pass on Windows?

zekth:
@mcollina i fixed the test. It seems that `tap` has changed its api for `t.match` which was matching a `Buffer` as an `Array`. I made a slight change to the test, it might not be pretty as the match pattern but i haven't found anything in the doc to match it properly

mcollina (approved):
lgtm

PR #222, opened 2021-04-19
Fix incorrect error types

### LINKED ISSUE #221: Wrong types of custom errors
## 🐛 Bug Report

Using `instanceof` to check if an error is a custom error of this library (such as `RequestFileTooLargeError`) results in a type error.

```ts
if (a instanceof fastify.multipartErrors.RequestFileTooLargeError) { ... }
```
```
The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type.
```

## To Reproduce

```ts
let a;
if (a instanceof fastify.multipartErrors.RequestFileTooLargeError) {
  // ignored
}
```

## Expected behavior

No type error.

## Fix

The types of the custom errors should be `FastifyErrorConstructor` instead of `FastifyError`, for they are not error instances, but constructors to create errors.

## Your Environment

- *node version*: 14
- *typescript version*: 4.2.3
- *fastify version*: >=3.0.0
- *os*: Windows

ddadaal:
Fix #221.

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,7 @@
 import * as busboy from "busboy";
 import { FastifyPlugin } from "fastify";
 import { Readable } from 'stream';
-import { FastifyError } from "fastify-error";
+import { FastifyErrorConstructor } from "fastify-error";
 
 type MultipartHandler = (
     field: string,
@@ -41,12 +41,12 @@ interface MultipartValue<T> {
 }
 
 interface MultipartErrors {
-    PartsLimitError: FastifyError,
-    FilesLimitError: FastifyError,
-    FieldsLimitError: FastifyError,
-    PrototypeViolationError: FastifyError,
-    InvalidMultipartContentTypeError: FastifyError,
-    RequestFileTooLargeError: FastifyError
+    PartsLimitError: FastifyErrorConstructor,
+    FilesLimitError: FastifyErrorConstructor,
+    FieldsLimitError: FastifyErrorConstructor,
+    PrototypeViolationError: FastifyErrorConstructor,
+    InvalidMultipartContentTypeError: FastifyErrorConstructor,
+    RequestFileTooLargeError: FastifyErrorConstructor
 }
 
 declare module "fastify" {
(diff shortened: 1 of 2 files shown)


test/index.test-d.ts
```diff
@@ -141,6 +141,14 @@ const runServer = async () => {
   // access all errors
   app.post('/upload/files', async function (req, reply) {
     const { FilesLimitError } = app.multipartErrors
+
+    // test instanceof Error
+    const a = new FilesLimitError();
```
mcollina, line 146: Can you use a `tsd` assertion?

ddadaal: It is used to test if an error type (like `FilesLimitError`) can be used in a `instanceof` expression. I am not sure how to do it.  Checking if the file compiles seems ok, but currently the file doesn't compile because of some other errors.

index.d.ts
```diff
@@ -41,12 +41,12 @@ interface MultipartValue<T> {
 }
 
 interface MultipartErrors {
-    PartsLimitError: FastifyError,
-    FilesLimitError: FastifyError,
-    FieldsLimitError: FastifyError,
-    PrototypeViolationError: FastifyError,
-    InvalidMultipartContentTypeError: FastifyError,
-    RequestFileTooLargeError: FastifyError
+    PartsLimitError: FastifyErrorConstructor,
```
climba03003, line 44: This is the wrong fixing approach. All the `MultiparErrors` is not construable. They should be the instance of `FastifyError` or `Error`.

climba03003 (changes requested):

test/index.test-d.ts
climba03003: You can simply adding this to the test file.
```ts
expectType<FastifyErrorConstructor>(FilesLimitError);
```

climba03003 (approved):
LGTM.

mcollina (approved):
lgtm

@climba03003 could you merge and release?

PR #248, opened 2021-06-29
Fix file stream type

### LINKED ISSUE #247: NodeJS.ReadableStream used over Readable
### Prerequisites

### Fastify version

3.x.x

### Plugin version

4.0.6

### Node.js version

16.3.0

### Operating system

Windows

10

### Description

According to https://nodejs.org/api/stream.html#stream_readable_wrap_stream:

```
When using an older Node.js library that emits 'data' events and has a stream.pause() method that is advisory only, the readable.wrap() method can be used to create a Readable stream that uses the old stream as its data source.

It will rarely be necessary to use readable.wrap() but the method has been provided as a convenience for interacting with older Node.js applications and libraries.
```

So usage of `NodeJS.ReadableStream` seems to be more of a legacy behaviour, and it causes issues when e. g. trying to pass this stream to Azure Storage, which expects to get Readable instead.

### Steps to Reproduce

Call req.file().file. See TS signature of fastify-multipart:

```
export interface MultipartFile {
  toBuffer: () => Promise<Buffer>,
  file: NodeJS.ReadableStream,
  filepath: string,
  fieldname: string,
  filename: string,
  encoding: string,
  mimetype: string,
  fields: MultipartFields
}
```

### Expected Behavior

Multipart file should be Readable

kibertoad:
#### Checklist

fixes #247

Current type is based on incorrect busBoy typings which will be addressed separately. In reality `Readable` is actually provided.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,7 +27,7 @@ export type Multipart<T = true> = T extends true ? MultipartFile : MultipartValu
 
 export interface MultipartFile {
   toBuffer: () => Promise<Buffer>,
-  file: NodeJS.ReadableStream,
+  file: Readable,
   filepath: string,
   fieldname: string,
   filename: string,
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -72,7 +72,7 @@ const runServer = async () => {
     expectError(req.body.foo.file);
     expectType<string>(req.body.foo.value);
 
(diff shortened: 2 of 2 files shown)


StarpTech:
LGTM

StarpTech (approved):

PR #259, opened 2021-07-28
Updates fileSize limits example in README

### LINKED ISSUE #250: limits.fileSize not working when greater than a specified value
### Prerequisites

### Fastify version

3.8.1

### Plugin version

4.0.7

### Node.js version

14.17.1

### Operating system

Linux

Manjaro 21

### Description

When fileSize is greater than a specified value, it will not throw exception when upload large file.

### Steps to Reproduce

Here is the sample code just from README

```js
const fastify = require('fastify')()
const fs = require('fs')
const util = require('util')
const path = require('path')
const { pipeline } = require('stream')
const pump = util.promisify(pipeline)

fastify.register(require('fastify-multipart'))

  // process a single file
  // also, consider that if you allow to upload multiple files
  // you must consume all files otherwise the promise will never fulfill
  console.log('before req file');
  // {limits: {fileSize: 1024}}
  const data = await req.file({limits: {fileSize: 40 * 1024}})
  console.log(data);
  console.log('after req file');

  // to accumulate the file in memory! Be careful!
  //
  // await data.toBuffer() // Buffer
  //
  // or

  console.log('before pump')
  await pump(data.file, fs.createWriteStream(data.filename))
  console.log('after pump')
  // be careful of permission issues on disk and not overwrite
  // sensitive files that could cause security risks
  
  // also, consider that if the file stream is not consumed, the promise will never fulfill

  reply.send()
})

fastify.listen(3000, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

Test with `curl`:

```sh
curl -v -F 'file=@BIG_FILE' localhost:3000/
```

When `fileSize` is a small one like `40 * 1024`(40KB), it's working, response:

```json
{"statusCode":413,"code":"FST_REQ_FILE_TOO_LARGE","error":"Payload Too Large","message":"request file too large, please check multipart config"}
```

But when `fileSize` is a large one like `40 * 1024 * 1024`(40MB), it's not working, the file is uploaded successfully, but the uploaded file is truncated to 40MB only.

### Expected Behavior

`limits.fileSize` should working as expect.

### LINKED ISSUE #256: limits.fileSize when set global fileSize doesn't get limited
### Prerequisites

### Fastify version

3.0.0

### Plugin version

4.0.5

### Node.js version

14.17.3

### Operating system

Linux

20.04

### Description

I'm setting up the plugin to limit to file size of 2 MB, it doesn't work this is my current setup
image

### Steps to Reproduce

As documented here `limit.fileSize` didn't work for me when set global(On plugin registration)

### Expected Behavior

The plugin should have limited to the size I've specified during its registration

leorossi:
Updates the example to work with `limits.fileSize` option.

Closes #256, Closes #250 

#### Checklist

- [X] documentation is changed or added

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,21 +90,18 @@ fastify.register(require('fastify-multipart'), {
 });
 ```
 
-If you do set upload limits, be sure to catch the error. An error or exception will occur if a limit is reached. These events are documented in more detail [here](https://github.com/mscdex/busboy#busboy-special-events).
-
 **Note**: if the file stream that is provided by `data.file` is not consumed, like in the example below with the usage of pump, the promise will not be fulfilled at the end of the multipart processing.
 This behavior is inherited from [busboy](https://github.com/mscdex/busboy).
 
 **Note**: if you set a `fileSize` limit and you want to know if the file limit was reached you can listen to `data.file.on('limit')` or check at the end of the stream the property `data.file.truncated`. 
 
 ```js
-try {
-  const data = await req.file()
-  await pump(data.file, fs.createWriteStream(data.filename))
-} catch (error) {
-  if (error instanceof fastify.multipartErrors.FilesLimitError) {
-    // handle error
-  }
+const data = await req.file()
+await pump(data.file, fs.createWriteStream(data.filename))
+if (data.file.truncated) {
+  // you may need to delete the part of the file that has been saved on disk
+  // before the `limits.fileSize` has been reached
+  reply.send(new fastify.multipartErrors.FilesLimitError());    
 }
 ``` 
 


simoneb (approved):

simoneb (approved):

README.md
```diff
@@ -90,21 +90,18 @@ fastify.register(require('fastify-multipart'), {
 });
 ```
 
-If you do set upload limits, be sure to catch the error. An error or exception will occur if a limit is reached. These events are documented in more detail [here](https://github.com/mscdex/busboy#busboy-special-events).
```
mcollina, line 93: why was this removed?

README.md
```diff
@@ -90,21 +90,18 @@ fastify.register(require('fastify-multipart'), {
...
 
 ```js
-try {
-  const data = await req.file()
-  await pump(data.file, fs.createWriteStream(data.filename))
-} catch (error) {
-  if (error instanceof fastify.multipartErrors.FilesLimitError) {
-    // handle error
-  }
+const data = await req.file()
```
StarpTech, line 99: LGTM. We should try to promote `const files = await req.saveRequestFiles()` higher than the lowe level api. As you already experience there are some edge cases. `saveRequestFiles()` is good for most of the use cases.

README.md
leorossi: Because the error is not thrown actually, check this comment: https://github.com/fastify/fastify-multipart/issues/250#issuecomment-888356293

simoneb:
@mcollina @StarpTech are we happy with this change? If so, can you please merge this PR?

mcollina (approved):
lgtm

Issue #277, opened 2021-10-01
Improve docs: add more details for JSON schema validation


radomird:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

When validating non-file fields in the body and **NOT** using the shared schema, according to the plugin documentation, we can validate the fields by using the following syntax:
```
body: {
  type: 'object',
  properties: {
    hello: {
      properties: {
        value: { 
          type: 'string',
        }
      }
    }
 }
```

However one would expect that for the non-file fields (like this for example: ```{ "hello": "world"}``` ) the JSON schema for validation would be:
```
body: {
  type: 'object',
  properties: {
    hello: {
      type: 'string'
    }
  }
}
```
just like it's mentioned in the [fastify docs](https://www.fastify.io/docs/latest/Validation-and-Serialization/).

However, the plugin converts the non-file fields from this:
 ```
{ "hello": "world"}
``` 
to this: 
```
{ 
  hello: {
    fieldname: "hello",
    value: "world",
    fieldnameTruncated: false,
    valueTruncated: false,
    fields: body
  }
}
```
and therefore the validation schema on top (from the plugin docs) makes sense.

However, **the problem is that this conversion is not mentioned anywhere** in the documentation and it's just really confusing and leads to issues like this: https://github.com/fastify/help/issues/525#issue-989998191

I would suggest to mention the conversion which would explain why is the validation schema defined like it is.

mcollina:
Would you like to send a PR?

radomird:
@mcollina Sure, will do that.

Issue #276, opened 2021-09-30
Update README with relevant information from callback.md file


radomird:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Hi,
I would like to propose a suggestion. In the Readme file there is some missing information regarding schema validation that is **_still relevant in the new API_** (and I think it should be there as it is very useful).

The missing information can be found at the end of the [callback.md](../blob/master/callback.md) file, I've pasted it below for reference:

> The sharedSchemaId parameter must provide a string ID and a shared schema will be added to your fastify instance so you will be able to apply the validation to your service like this:

```
fastify.post('/upload', {
  schema: {
    body: {
      type: 'object',
      required: ['myStringField', 'myFilenameField'],
      properties: {
        myStringField: { type: 'string' },
        myFilenameField: { type: 'array', items: 'MultipartFileType#' }
    }
  }
}, function (req, reply) {
  reply.send('done')
})
```

> The shared schema added will be like this:

```
{
  type: 'object',
  properties: {
    encoding: { type: 'string' },
    filename: { type: 'string' },
    limit: { type: 'boolean' },
    mimetype: { type: 'string' }
  }
}
```

radomird:
> The readme contains
> 
> > If you enable attachFieldsToBody and set sharedSchemaId a shared JSON Schema is added, which can be used to validate parsed multipart fields.
> 
> how could we write it clearer?

That part is really clear, but the information that I quoted from callback.md file is actually missing from the readme.

mcollina:
Would you like to send a PR?

radomird:
@mcollina Sure, will do that.

PR #280, opened 2021-10-04
Updated README.md with relevant information about the sharedSchemaId

radomird:
Updated the README with additional (missing) information that is relevant for `sharedSchemaId`. 
Closes: https://github.com/fastify/fastify-multipart/issues/276, closes https://github.com/fastify/fastify-multipart/issues/277

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,16 @@ The mentioned field will be converted, by the plugin, to a more complex field un
 }
 ```
 
-It is important to know that this conversion happens BEFORE the field is validated, so keep that in mind when writing the JSON schema for validation for fields that don't use the shared schema.
+It is important to know that this conversion happens BEFORE the field is validated, so keep that in mind when writing the JSON schema for validation for fields that don't use the shared schema. The schema for validation for the field mentioned above should look like this:
+```js
+hello: {
+  properties: {
+    value: { 
+      type: 'string'
+    }
+  }
+}
+```
 
 ## Access all errors
 
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ fastify.post('/upload/files', {
       properties: {
         // field that uses the shared schema
         myField: { $ref: '#mySharedSchema'},
-        // or another field that uses shared schema
+        // or another field that uses the shared schema
         myFiles: { type: 'array', items: fastify.getSchema('mySharedSchema') },
         // or a field that doesn't use the shared schema
         hello: {
--- a/README.md
+++ b/README.md
@@ -293,6 +293,21 @@ fastify.post('/upload/files', {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

README.md
```diff
@@ -293,6 +294,42 @@ fastify.post('/upload/files', {
...
+}
+```
+**Note**:
+When sending fields with the body (`attachFieldsToBody` set to true), a field might look like this:
+```json
+{
+  "hello": "world"
+}
+```
+The mentioned field will be converted, by the plugin, to a more complex field under the hood. The converted field will look something like this:
```
Eomm, line 318: This is not totally clear to me reading the docs what is the trigger that transform the `hello` field into a complex one

If `attachFieldsToBody:true` we will get `request.body` as `{ "hello": "world"}`
If "something else" we will get the `request.body.hello` as a complex field

radomird: Actually the plugin will ALWAYS convert ALL the non file fields.

Eomm: Ok, so we are saying that the `hello: world` is the expectation, but the user will get the complex one.
I would remove this example, it carries the attention to what the user will not have in its hands.

I would just say:

> Plain fields' JSON structure is like: ..

And I would put a JSON schema near the example, this would be really helpful for users

radomird: Well, the whole reason of bringing this to the users attention is that when adding a the JSON schema for the plain field would be expected to be:
```js
hello: {
  type: 'string'
}
```
But instead (because of the conversion that is NOT mentioned anywhere) the user has to write it like this:
```js
hello: {
  properties: {
    value: { 
      type: 'string',
    }
  }
 }
```
If we don't mention the conversion and the result of it - regardless of the fact that it's not in the users hands, then the schema will not make sense or it will be confusing (like it was for me before).

radomird: I've added the JSON schema bit under the example.

README.md
```diff
@@ -293,6 +294,42 @@ fastify.post('/upload/files', {
...
+  properties: {
+    encoding: { type: 'string' },
+    filename: { type: 'string' },
+    limit: { type: 'boolean' },
+    mimetype: { type: 'string' }
+  }
+}
+```
+**Note**:
+When sending fields with the body (`attachFieldsToBody` set to true), a field might look like this:
```
Eomm, line 312:
```suggestion
When sending fields with the body (`attachFieldsToBody` set to true), the field might look like this into the `request.body`:
```

radomird: Did you mean this:
```
When sending fields with the body (`attachFieldsToBody` set to true), the field might look like this in the `request.body`:
```
?

Eomm: Yes, thks 😄

Eomm (approved):

PR #305, opened 2021-12-08
Fix TS types

### LINKED ISSUE #304: Multipart: Boundary not found
### Prerequisites

### Fastify version

3.24.1

### Plugin version

5.2.0

### Node.js version

16.5.0

### Operating system

Linux

Ubuntu 20.04

### Description

When the file limit size is set, then headers are a mandatory prop. The code should be something like:

```js
await req.file({
    headers: { "content-type": "multipart/form-data" },
    limits: {
        fileSize: (1024 * 1024) * 55
    }
})
```

The issue begins with passing those content-type at the request (as they are mandatory now), then the server responds with error 500 and the message "Multipart: Boundary not found".

### Steps to Reproduce

1 - Make a file upload;
2 - Set some limit to the upload throught req.file(PROPS);
3 - Verify that the headers are mandatory now;
4 - Pass the respective header throught request;
5 - Try to do the upload;
6 - Observe that the server responds with error 500 with the message "Multipart: Boundary not found".

### Expected Behavior

The upload happens normally respecting the file size limit.

kibertoad:
fixes #304

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,17 +54,17 @@ declare module "fastify" {
         isMultipart: () => boolean;
 
         // promise api
-        parts: (options?: Partial<BusboyConfig>) =>  AsyncIterableIterator<Multipart>
+        parts: (options?: Omit<BusboyConfig, 'headers'>) =>  AsyncIterableIterator<Multipart>
 
         // legacy
-        multipart: (handler: MultipartHandler, next: (err: Error) => void, options?: Partial<BusboyConfig>) => Busboy;
+        multipart: (handler: MultipartHandler, next: (err: Error) => void, options?: Omit<BusboyConfig, 'headers'>) => Busboy;
 
         // Stream mode
-        file: (options?: Partial<BusboyConfig>) => Promise<Multipart>
-        files: (options?: Partial<BusboyConfig>) => AsyncIterableIterator<Multipart>
+        file: (options?: Omit<BusboyConfig, 'headers'>) => Promise<Multipart>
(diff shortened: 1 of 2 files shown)


kibertoad:
@climba03003 Technically busboy types are correct, it does need that parameter for functioning. However, fastify-multipart types were wrong, because we were not relying on user to pass that parameter, we were retrieving it from request directly.

index.d.ts
```diff
@@ -90,6 +90,16 @@ export interface FastifyMultipartBaseOptions {
...
+    /**
+     * Detect if a Part is a file.
+     *
+     * By default a file is detected if contentType
+     * is application/octet-stream or fileName is not
+     * undefined.
+     *
+     * Modify this to handle e.g. Blobs.
+     */
+    isPartAFile?: (fieldName: string | undefined, contentType: string | undefined, fileName: string | undefined) => boolean;
```
kibertoad, line 102: This part of busboy types was missing

climba03003:
> @climba03003 Technically busboy types are correct, it does need that parameter for functioning. However, fastify-multipart types were wrong, because we were not relying on user to pass that parameter, we were retrieving it from request directly.

I will use `Omit<BusboyConfig, 'headers'>` as we handle this already. 
`Partial` also allow the users to pass the `headers` which may lead to the same problem again.

kibertoad:
@climba03003 Good point. Fixed.

climba03003 (approved):
LGTM.

rhadamez:
You guys are too fast, thats how a real framework should happen. Thanks.

Issue #315, opened 2022-01-03
Canceled uploads are not propagated to the part.file FileReadStream


mccare:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.25.3

### Plugin version

5.2.1

### Node.js version

16.13.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.1

### Description

When a user cancels the upload request, the cancel/destroy event is not propagated to the `part.file` FileReadStream. The request is not ended. Cleanup code that lives e.g. in the `finally` of my `try/catch` block is not called (e.g. to remove tempfiles).

### Steps to Reproduce

I've created a sample repo here https://github.com/mccare/multipart-upload-error-handling

* yarn install and run yarn dev
* `curl -i --limit-rate 100k -w '%{http_code}\n' -F video=@bigfile.binary -F events=@anotherbigfile.binary http://localhost:3000/upload `
* Terminate curl before request has gone through

The log files will show that the `finally` block is never reached. I'm assuming the request is not finished

### Expected Behavior

The code should throw some sort of error and the execution should at least reach the `finally` clause of my request handler.

I have added a `close` handler for `request.raw` to cancel/destroy the `part.file` Readable. This produces the expected results, but I don't know if that is the correct way to handle this.

climba03003:
@Uzlopak do you have time to look at this issue? I think it is something related to both `busboy` and how request is ended.

Uzlopak:
Hi, i arrived just few hours ago in Germany, where my Laptop is. I will have a look at it.

My assessment is close to yours. I think BusBoy-Stream has to listen to the close event and pass it down to each open partsStram.

Uzlopak:
@climba03003 @mccare 
I checked it out. request.raw is calling unpipe on busboy
https://github.com/fastify/fastify-multipart/blob/635a523bdf1ae42bf8aa2191d6f10f11665e5930/index.js#L472

So it should make the trick to modify the code about here https://github.com/fastify/busboy/blob/6d8f8741cd5a9787fe7c4d1f1fb57bd165ad6c66/lib/types/multipart.js#L174 by adding in line 175
```javascript
        boy.on('unpipe', function () {
          file.destroy()
        })
```
Maybe also adding this makes sense.
```javascript
        boy.on('cancel', function () {
          file.destroy()
        })
        boy.on('destroy', function () {
[...]

PR #319, opened 2022-01-17
feat: added destroying of part.file if the node request ends with an error event

mccare:
If the client cancels the http request (e.g. just closes the socket),
the underlying file stream will not be closed with an error but
the request handling will hang.

The tests saves the stream of a file part into a Noop write stream
through pipeline. If the client cancels the request, the pipeline
call will not return. With the fix, we keep track of the file that
is currently being processed and in the case of an error event on the
node request, this file will be destroyed.

This addresses #315 and makes https://github.com/fastify/busboy/pull/81 superflous.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -316,6 +316,7 @@ function fastifyMultipart (fastify, options, done) {
 
     const body = {}
     let lastError = null
+    let currentFile = null
     const request = this.raw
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

zekth (approved):

Issue #252, opened 2021-07-02
other fields lost when upload a large file


abcfy2:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure it has not already been reported

### Fastify version

3.8.1

### Plugin version

4.0.7

### Node.js version

14.17.1

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Manjaro 21

### Description

When upload a large file. the fields will be lost.

### Steps to Reproduce

Here is the sample code:

```js
  const data = await req.file();
  const fields = data.fields;
  console.log(fields);
```

When upload a large file like:

```sh
curl -vF 'file=@BIG_FILE' -F 'name=n1' -F 'description=d1' -F 'tokenId=222' -F 'attributes=["v1","v2","v3"]' localhost:3000/
```

The output is:

```sh
<ref *1> {
  file: {
    fieldname: 'file',
    filename: 'BIG_FILE.exe',
    encoding: '7bit',
    mimetype: 'application/octet-stream',
    file: FileStream {
      _readableState: [ReadableState],
      _events: [Object: null prototype],
      _eventsCount: 2,
      _maxListeners: undefined,
      truncated: false,
      _read: [Function (anonymous)],
      [Symbol(kCapture)]: false
    },
    fields: [Circular *1],
    _buf: null,
    toBuffer: [AsyncFunction: toBuffer]
  }
}
```

There is only `file` field, others like `name`, `description` lost.

mcollina:
Can you provide steps to reproduce? We need a full, runnable example.

abcfy2:
Here is the sample `index.js`:

```js
const fastify = require('fastify')()
const fs = require('fs')
const util = require('util')
const path = require('path')
const { pipeline } = require('stream')
const pump = util.promisify(pipeline)

fastify.register(require('fastify-multipart'))

fastify.post('/', async function (req, reply) {
  const data = await req.file()
  const fields = data.fields
  console.log(fields)
  console.log(fields.k1)
  console.log(fields.k2)
  await pump(data.file, fs.createWriteStream('tmp_file'))
  reply.send()
})

fastify.listen(3000, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

We also create two files, one is small, another is big:

```sh
$ fallocate -l 10KB SMALL_FILE
$ fallocate -l 100MB BIG_FILE
```

[...]

PR #320, opened 2022-01-19
docs: move data.fields line in README

frapan:
#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ fastify.listen(3000, err => {
 
 **Note** about `data.fields`: `busboy` consumes the multipart in serial order (stream). Therefore, the order of form fields is *VERY IMPORTANT* to how `fastify-multipart` can display the fields to you.
 We would recommend you place the value fields first before any of the file fields.
-It can ensure your fields will be accessible before start consuming any file.
+It will ensure your fields are accessible before it starts consuming any files.
 If you cannot control the order of the placed fields, be sure to read `data.fields` *AFTER* consuming the stream, or it will only contain the fields parsed at that moment.
 
 You can also pass optional arguments to `@fastify/busboy` when registering with Fastify. This is useful for setting limits on the content that can be uploaded. A full list of available options can be found in the [`@fastify/busboy` documentation](https://github.com/fastify/busboy#busboy-methods).
--- a/README.md
+++ b/README.md
@@ -72,9 +72,6 @@ fastify.listen(3000, err => {
 
 You can also pass optional arguments to `@fastify/busboy` when registering with Fastify. This is useful for setting limits on the content that can be uploaded. A full list of available options can be found in the [`@fastify/busboy` documentation](https://github.com/fastify/busboy#busboy-methods).
 
-**Note**: if the file stream that is provided by `data.file` is not consumed, like in the example below with the usage of pump, the promise will not be fulfilled at the end of the multipart processing.
-This behavior is inherited from [`@fastify/busboy`](https://github.com/fastify/busboy).
-
 ```js
 fastify.register(require('fastify-multipart'), {
   limits: {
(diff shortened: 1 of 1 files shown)


frapan:
Fixes #252

climba03003 (changes requested):
`(available only AFTER the stream is consumed)`
The above sentence is not always true. I do not want to give wrong explanation to the client just to reduce the issue.

Instead, I would like someone to explain more on how `stream` or `busboy` operate and create a Q&A section.

e.g.
`busboy` consume the `multipart` in serial order (stream), the **order** of form field is `VERY IMPORTANT` to how `fastify-multipart` can display the fields to you. We would recommend you to place the value fields first before any of the file fields. It can ensure your `fields` will be accessible before start consuming any file.

darkgl0w (changes requested):
> Note: if the file stream that is provided by data.file is not consumed, like in the example below with the usage of pump, the promise will not be fulfilled at the end of the multipart processing. This behavior is inherited from @fastify/busboy.

There is a note about this in the `README.md` just after the example

I would suggest that you add another note with the explaining sentence proposed by @climba03003 instead.

frapan:
Thank you for your explanations.

I reverted to the original file and added the note of @climba03003 with a small addition:

> **Note** about `data.fields`: `busboy` consumes the multipart in serial order (stream). Therefore, the order of form fields is *VERY IMPORTANT* to how `fastify-multipart` can display the fields to you.
> We would recommend you to place the value fields first before any of the file fields.
> It can ensure your fields will be accessible before start consuming any file.
> If you cannot control the order of the placed fields, be sure to read `data.fields` *AFTER* consuming the stream, or it will contain only the fields parsed at that moment.

I also realized that the following note was duplicated:

> Note: if the file stream that is provided by data.file is not consumed, like in the example below with the usage of pump, the promise will not be fulfilled at the end of the multipart processing. This behavior is inherited from @fastify/busboy.

So I deleted the one that was in the wrong (imho) position.

What do you think?

mccare:
On a related note, the documentation here https://github.com/fastify/busboy#busboy-special-events mentions that you can also just use `file.resume()`. If that is correct, I would also add it to the documentation here. WDYT?

climba03003:
> On a related note, the documentation here https://github.com/fastify/busboy#busboy-special-events mentions that you can also just use `file.resume()`. If that is correct, I would also add it to the documentation here. WDYT?

I think it is a separate issue.

climba03003:
Can you kindly help to check the sentence for us?
cc @Fdawgs

climba03003 (approved):
LGTM.

README.md
```diff
@@ -70,10 +70,12 @@ fastify.listen(3000, err => {
 })
 ```
 
-You can also pass optional arguments to `@fastify/busboy` when registering with Fastify. This is useful for setting limits on the content that can be uploaded. A full list of available options can be found in the [`@fastify/busboy` documentation](https://github.com/fastify/busboy#busboy-methods).
+**Note** about `data.fields`: `busboy` consumes the multipart in serial order (stream). Therefore, the order of form fields is *VERY IMPORTANT* to how `fastify-multipart` can display the fields to you.
+We would recommend you to place the value fields first before any of the file fields.
```
Fdawgs, line 74:
```suggestion
We would recommend you place the value fields first before any of the file fields.
```

README.md
```diff
@@ -70,10 +70,12 @@ fastify.listen(3000, err => {
...
+It can ensure your fields will be accessible before start consuming any file.
```
Fdawgs, line 75:
```suggestion
It will ensure your fields are accessible before it starts consuming any files.
```

README.md
```diff
@@ -70,10 +70,12 @@ fastify.listen(3000, err => {
...
+If you cannot control the order of the placed fields, be sure to read `data.fields` *AFTER* consuming the stream, or it will contain only the fields parsed at that moment.
```
Fdawgs, line 76:
```suggestion
If you cannot control the order of the placed fields, be sure to read `data.fields` *AFTER* consuming the stream, or it will only contain the fields parsed at that moment.
```

Fdawgs (changes requested):

mcollina:
@Fdawgs are you ok now?

Fdawgs (approved):

Fdawgs:
> @Fdawgs are you ok now?

All gravy now

PR #348, opened 2022-06-03
Option to attach to body as key-value pairs

zone117x:
Adds an option `attachFieldsToBody: 'valueOnly'`. When this is option is enabled, the field values are attached to the body directly, similar to what is described in the readme:
```js
  const body = Object.fromEntries(
    Object.keys(req.body).map((key) => [key, req.body[key].value])
  ) // Request body in key-value pairs, like req.body in Express
```

This would be a convenient feature to include so that json-schemas can be shared more easily between the content types `application/json`, `application/x-www-form-urlencoded`, and `multipart/form-data`. 

Without this capability, this needs to be performed in awkward route handler transforms, e.g.:
```ts
fastify.post<{ Body: ServiceAuthRequest, Reply: ServiceAuthResponse }>('/request_key', {
  preValidation: (request, _reply, done) => {
    if (request.isMultipart()) {
      // 🤢
      const body = request.body as unknown as Record<string, { value: string } | { _buf: Buffer }>;
      for (let [k, v] of Object.entries(body)) {
        const value = 'value' in v ? v.value : v._buf.toString();
        (request.body as any)[k] = value;
      }
    }
    done();
  },
  schema: {
    body: ServiceAuthRequestSchema,
    headers: ServiceAuthRequestHeadersSchema,
    response: {
      200: ServiceAuthResponseSchema,
    }
  },
}, 
async (request, reply) => {
  ...
});
```

---

Please let me know if the the convention `attachFieldsToBody: 'valueOnly'` should be changed to something else.

Also, currently files are converted to a string via `.toBuffer().toString()`. It's possible to use the `onFile` handler to customize the file buffer decoding, which is performed in one of the new tests.

Let me know if this feature make sense, and I'm happy to add info to the docs/readme.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,17 @@ fastify.post('/upload/files', async function (req, reply) {
 })
 ```
 
+Request body key-value pairs can be assigned directly using `attachFieldsToBody: 'keyValues'`. Field values will be attached directly to the body object. By default, all files are converted to a string using `buffer.toString()` used as the value attached to the body.
+
+```js
+fastify.register(require('@fastify/multipart'), { attachFieldsToBody: 'keyValues' })
+
+fastify.post('/upload/files', async function (req, reply) {
+  const uploadValue = req.body.upload // access file as string
+  const fooValue = req.body.foo       // other fields
+})
+```
+
 You can also define an `onFile` handler to avoid accumulating all files in memory.
 
 ```js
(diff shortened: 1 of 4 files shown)


kibertoad:
This makes sense to me.
@uzlopak wdyt?

Uzlopak:
It looks kind of confusing that valuesOnly still creates key value pairs. Also is the for loop fast? I would do classical with Object.keys.
And lastly does it not potentially needs protection against prototype pollution?

zone117x:
> And lastly does it not potentially needs protection against prototype pollution?

On second look, I don't think additional changes are need for this. The request body is using the same set of keys, and only re-assigning the value. Prototype pollution should already be handled for these keys by the time this path is reached, right?

index.js
```diff
@@ -147,6 +147,18 @@ function fastifyMultipart (fastify, options, done) {
...
         }
       }
+      if (options.attachFieldsToBody === 'keyValues') {
+        const body = {}
+        for (const key of Object.keys(req.body)) {
+          const field = req.body[key]
+          if (field.value !== undefined) {
+            body[key] = field.value
+          } else if (field._buf !== undefined) {
+            body[key] = field._buf.toString()
```
climba03003, line 157: I don't think it is a good idea for `.toString()` a buffer directly.

Uzlopak:
I think it is now good.

index.js
climba03003: Since, it is `buffer` and people usually upload photo using `form`.
I would probably encode it as `base64`.

climba03003 (changes requested):
Documentation is required.

index.js
Uzlopak: I dont know if this is the best way to do it... either keep it as a buffer or encode it to a binary string. Encoding to a base64 string would just result 30% memory overhead and you would still need to decode it base64 to make something useful with it.

climba03003: Keep it as `buffer` will invalidate by the `validation` phase which do not match the goal of this PR.
Put the whole file buffer inside memory is not a good deal already. I would accept even more consumption in this case.

For `binary` string, it can be an option. But it needs to document how to transform it back to `buffer` or how to use it with `stream` or `fs.writeFile`, etc.

zone117x: > people usually upload photo using form

It seems like for a photo upload use-case, the `attachFieldsToBody` option wouldn't be used, as it buffers the entire photo in memory which could be quite large. If people still want to do that, then can use `attachFieldsToBody: true`. This PR does not _remove_ that existing option so it will no have no impact on that use case.

A use-case that I'd like to support is curl commands which can include key values directly in the command, as well as point to files for the value, e.g.

```shell
curl -i -X POST http://localhost:1234/auth \
  -F "username=matt" \
  -F "key=@privatekey.txt" \
  -F "thing=123"
```

Additionally, the `buffer.toString()` behavior can be bypassed / customized using the existing `onFile` handler. I've included an example of using it to decode base64 in the unit tests:

https://github.com/fastify/fastify-multipart/blob/8eaf5f303ea59be087b5dbeb62b40b42894c9c8e/test/multipart-attach-body.test.js#L115-L120

zone117x:
Documentation added for the new `attachFieldsToBody: 'valueOnly'` option, and also how it improves JSON Schema validation.

Eomm (approved):

README.md
```diff
@@ -250,13 +261,60 @@ fastify.post('/upload/files', async function (req, reply) {
...
+```
+
 **Note**: if you assign all fields to the body and don't define an `onFile` handler, you won't be able to read the files through streams, as they are already read and their contents are accumulated in memory.
 You can only use the `toBuffer` method to read the content.
 If you try to read from a stream and pipe to a new file, you will obtain an empty new file.
 
 ## JSON Schema body validation
 
-If you enable `attachFieldsToBody` and set `sharedSchemaId` a shared JSON Schema is added, which can be used to validate parsed multipart fields.
+If you enable `attachFieldsToBody: 'keyValues'` then the response body and JSON Schema validation will behave the similary to `application/json` and [`application/x-www-form-urlencoded`](https://github.com/fastify/fastify-formbody) content types. Files will be decoded using `Buffer.toString()` and attached as a body value.
```
kibertoad, line 287:
```suggestion
If you enable `attachFieldsToBody: 'keyValues'` then the response body and JSON Schema validation will behave similarly to `application/json` and [`application/x-www-form-urlencoded`](https://github.com/fastify/fastify-formbody) content types. Files will be decoded using `Buffer.toString()` and attached as a body value.
```

climba03003 (approved):

kibertoad:
Thank you!

savvykir:
Any updates in here?

PR #352, opened 2022-06-08
add .npmrc with package-lock=false to disable package-lock.json generation

Uzlopak:
#### Checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
climba03003, line 1: Why it is needed? and what does it solve?

climba03003: I means why it need to disable the lock file generation.
If it is something specific to your environment or personal perference, then it shouldn't be checked in git repository.

Uzlopak: Main Repo has the .npmrc also
https://github.com/fastify/fastify/blob/main/.npmrc

climba03003: That's not a reason.
Here is the PR adding `.npmrc` with other files in `fastify`. 
https://github.com/fastify/fastify/pull/2269

But, it didn't explain why disable the `package-lock.json` improve DX.
We can use `npm udpate` if we install the dependency once.

Uzlopak: npm update does not depend on the lockfile.

@jsumners Any argument why disabling the unnecessary generation of the package-lock.json is a good idea?

climba03003: Disable `package-lock.json` means every run for `npm install` should fetch and install the latest.
Which means after the first `npm install`, it is actually doing the same thing as `npm update`.

jsumners (approved):

.npmrc
jsumners: There is not point to having a package lock in a library project. They are ignored when installing the module as a dependency. The only thing having a package lock in this project will accomplish is to lock versions across contributors to the time when the various contributors start their work on the project.

PR #353, opened 2022-06-08
add concat-stream to devDependencies

Uzlopak:
Currently, rimraf and concat-stream are imported because they are dependencies of other packages. This PR adds them explicitly.


climba03003 (approved):
LGTM.

Can we replace `rimraf` with `fs.rmdirSync`

kibertoad:
@climba03003 I doubt it, if they are transitive dependencies, unless we send PRs upstream.
@Uzlopak do you know which packages depend on rimraf?

climba03003:
> I doubt it, if they are transitive dependencies, unless we send PRs upstream.

I means it is added in `devDepenencies` because we directly using it.
If we remove the direct usage, then it can be removed in `devDepenencies`.

Uzlopak:
@kibertoad 
tap requires rimraf, as it is a dependency of nyc.

climba03003:
> What do you prefer?

`await fs.promises.rm(tmpdir, { recursive: true, force: true })`

Since we always test against the latest version of node (inside that major).
If all of them provide `fs.promises.rm`, then no need a extra check for it.

climba03003 (approved):
Thanks

kibertoad (approved):

PR #359, opened 2022-06-15
Create key values when multiple files are uploaded

vasilianeugen:
When you define sharedSchemaId and you want to allow multiple files upload, the files are in array and no value can be returned from onFile in this case.


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

PR #390, opened 2022-09-13
Improve typescript types

### LINKED ISSUE #251: Typescript data.fields returns wrong type.
### Prerequisites

### Fastify version

3.8.1

### Plugin version

4.0.7

### Node.js version

14.17.1

### Operating system

Linux

Manjaro 21

### Description

Seems that the type define of `fields` is not correct.

```ts
export type Multipart<T = true> = T extends true ? MultipartFile : MultipartValue<T>;
```

Seems this make `Multipart` type is always `MultipartFile`, never `MultipartValue`.

### Steps to Reproduce

image
image

image

Building error:

```sh
TS2339: Property 'value' does not exist on type 'MultipartFile | MultipartFile[]'.
  Property 'value' does not exist on type 'MultipartFile'.
```

### Expected Behavior

`data.fields.{field_name}` should able to return `MultipartValue` type.

dwickern:
I've tried to correct the types by analyzing the source code and busboy's type definitions.

Fixes #251

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,7 +20,7 @@ interface BodyEntry {
 }
 
 export interface MultipartFields {
-    [fieldname: string]: Multipart | Multipart[];
+    [fieldname: string]: Multipart | Multipart[] | undefined;
 }
 
 export type Multipart = MultipartFile | MultipartValue;
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ import { FastifyErrorConstructor } from "@fastify/error";
 
 type MultipartHandler = (
     field: string,
-    file: any,
+    file: BusboyFileStream,
     filename: string,
     encoding: string,
     mimetype: string,
@@ -20,15 +20,14 @@ interface BodyEntry {
 }
 
 export interface MultipartFields {
-    [x: string]: Multipart | Multipart[];
+    [fieldname: string]: Multipart | Multipart[];
 }
 
-export type Multipart<T = true> = T extends true ? MultipartFile : MultipartValue<T>;
+export type Multipart = MultipartFile | MultipartValue;
 
 export interface MultipartFile {
   toBuffer: () => Promise<Buffer>,
   file: BusboyFileStream,
-  filepath: string,
   fieldname: string,
   filename: string,
   encoding: string,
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -20,24 +20,36 @@ interface BodyEntry {
...
-  value: T
+export interface SavedMultipartFile extends MultipartFile {
+  /**
+   * Path to the temporary file
+   */
+  filepath: string,
+}
+
+export interface MultipartValue<T = unknown> {
+  value: T;
```
dwickern, line 46: The value should be `unknown`, since it might be parsed as JSON based on the content-type provided by the caller. I've left it generic in case someone wants to assert its type.

index.d.ts
```diff
@@ -20,24 +20,36 @@ interface BodyEntry {
...
+  fieldname: string;
+  mimetype?: string;
```
dwickern, line 48: `string | undefined` because mime type get swallowed when we parse the value as JSON. I think this is a bug, but that's for another PR.

mcollina (approved):
lgtm

leppaott:
Our working code wasn't valid after this:
`await part.toBuffer();`

Adding:
```
if ('file' in part) {
            imageBuffer = await part.toBuffer();
```

according to those unit tests this became valid but it wasn't at all obvious based on types.
            
Don't think the  README example: https://github.com/fastify/fastify-multipart#handle-multiple-file-streams-and-fields

```
const parts = req.parts()
    if (part.file) {
      await pump(part.file, fs.createWriteStream(part.filename))
    } else {
      console.log(part)
    }
  }
```

will work anymore because part.file isn't available unless you do 'file' in part so IMO should revisit this.

test/index.test-d.ts
```diff
@@ -109,10 +123,10 @@ const runServer = async () => {
   app.post('/upload/raw/any', async function (req, reply) {
     const parts = req.parts()
     for await (const part of parts) {
-      if (part.file) {
+      if ('file' in part) {
```
leppaott, line 126: This should be modified on README as well. But this needs to be revisited to be on the type.

climba03003: Your problem is `TypeScript` issue.
`TypeScript` cannot restrict the type in certain format that's the cause.

The `README` example is certainly valid in `Javascript`.

dwickern: Yes, typescript wants you to use `in` for narrowing in this situation, which works for both JS and TS.
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing

A discriminator would perhaps be ideal:
```ts
export interface MultipartFile {
  type: 'file',
  ...
}
export interface MultipartValue<T = unknown> {
  type: 'value',
  ...
}

if (part.type === 'file') {
  part.toBuffer()
}
```

PR #391, opened 2022-09-14
Always include form value's content-type

dwickern:
Include the content-type even if it's not JSON.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -372,8 +372,6 @@ function fastifyMultipart (fastify, options, done) {
     request.pipe(bb)
 
     function onField (name, fieldValue, fieldnameTruncated, valueTruncated, encoding, contentType) {
-      let mimetype
-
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -390,7 +388,6 @@ function fastifyMultipart (fastify, options, done) {
 
         try {
           fieldValue = secureJSON.parse(fieldValue)
-          mimetype = 'application/json'
```
Eomm, line 393: I would keep this line to release this change as feature instead of breaking:

if users are using `mimetype === 'application/json'` and the content type is `application/json+vnd` we don't want to mutate the value

dwickern: Sure. It would also affect `application/json;charset=utf-8` which some applications incorrectly send.

mcollina (approved):
lgtm

Issue #313, opened 2021-12-21
Typescript: there is no way to make req.saveRequestFiles() works when we want to validate schema


meotimdihia:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I have multi problems with the plugin:
1. I am getting the error: typescript error: `Type 'never' must have a '[Symbol.iterator]()' method that returns an iterator. for this`
2. await req.saveRequestFiles() doesn't work unless I remove all validations and configs: `attachFieldsToBody`, `sharedSchemaId`, `onFile`
3. 

```Typescript
// config
fastify.register(fastifyMultipart, {
	attachFieldsToBody: true,
	sharedSchemaId: "#mySharedSchema",
	onFile
})

//routes
  const body = {
    type: "object",
    properties: {
      images: {
            type: "array",
            items: fastify.getSchema("mySharedSchema")
      }
    },
    additionalProperties: true
  } as const

  fastify.post<{
    Params: FromSchema<typeof params>
    Body: FromSchema<typeof body>
  }>("/upload",
    {
      schema: {
        body: body,
        response: {
          200: {
            type: "object",
            additionalProperties: true
          }
        }
      }
    },
    async (req, reply) => {
     // typescript error: Type 'never' must have a '[Symbol.iterator]()' method that returns an iterator.
      for (const file of req.body.images) {

      }  
      
      // doesn't work
      const files = await req.saveRequestFiles()
      
      // doesn't work
      req.files
      
```

Eomm:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

cc @fastify/typescript

meotimdihia:
@Eomm I am not really too good about Typescript.
I don't understand what is this error. 😀

RafaelGSS:
@meotimdihia Could you provide a minimal reproducible code? Further information, see: https://stackoverflow.com/help/minimal-reproducible-example

meotimdihia:
@RafaelGSS @Eomm Please re-open this issue.

This is the minimal reproducible code
https://github.com/meotimdihia/fastify-multipart-bug-report

### STEP 1 FAILED

When you upload files you will get this response `[]`

### STEP 2 OK

If you remove `attachFieldsToBody` in `app.ts` and remove `schema: { body: body}` in `root.ts`. When you upload files , you will get this response.

[...]

PR #409, opened 2023-01-09
Support for saveRequestFiles with attachFieldsToBody set true

Ceres6:
Added files async files generator function to `saveRequestFiles` which fetches files from body instead of from `getMultipartFiles` when `attachFieldsToBody is set to true

Partially addresses #313

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ const util = require('util')
 const createError = require('@fastify/error')
 const sendToWormhole = require('stream-wormhole')
 const deepmergeAll = require('@fastify/deepmerge')({ all: true })
-const { PassThrough, pipeline } = require('stream')
+const { PassThrough, pipeline, Readable } = require('stream')
 const pump = util.promisify(pipeline)
 const secureJSON = require('secure-json-parse')
 
@@ -101,6 +101,7 @@ function busboy (options) {
 }
 
 function fastifyMultipart (fastify, options, done) {
+  const attachFieldsToBody = options.attachFieldsToBody
   if (options.addToBody === true) {
     if (typeof options.sharedSchemaId === 'string') {
       fastify.addSchema({
@@ -139,6 +140,7 @@ function fastifyMultipart (fastify, options, done) {
       }
       for await (const part of req.parts()) {
         req.body = part.fields
+
         if (part.file) {
           if (options.onFile) {
             await options.onFile(part)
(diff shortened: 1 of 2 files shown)


mcollina:
Why does it only partially address https://github.com/fastify/fastify-multipart/issues/313?

mcollina (approved):
lgtm

Ceres6:
> Why does it only partially address #313?

Because the typescript problem described couldn't be reproduced, and in case of custom `onFile` it will only work as long as it still creates the `_buf` property on files attached to body. I guess that if that's not the case the user should as well store the files in that method as well.

mcollina:
> and in case of custom onFile it will only work as long as it still creates the _buf property on files attached to body.

Could you throw a proper error if this is the case?

index.js
```diff
@@ -34,30 +34,41 @@ function attachToBody (options, req, reply, next) {
...
-        file.destroy(err)
+  const mp = req.multipart(
+    (field, file, filename, encoding, mimetype) => {
+      body[field] = body[field] || []
+      body[field].push({
+        data: [],
+        filename,
+        encoding,
+        mimetype,
+        limit: false
```
mcollina, line 45: is this change relevant?

index.js
```diff
@@ -34,30 +34,41 @@ function attachToBody (options, req, reply, next) {
...
+      }
+    },
+    function (err) {
+      if (!err) {
+        req.body = body
+      }
+      next(err)
+    },
+    options
+  )
```
mcollina, line 71: is this change relevant? Seems just linting

index.js
```diff
@@ -77,8 +88,14 @@ function attachToBody (options, req, reply, next) {
...
-  file.on('data', data => { if (!lastFile.limit) { fileData.push(data) } })
-  file.on('limit', () => { lastFile.limit = true })
+  file.on('data', (data) => {
+    if (!lastFile.limit) {
+      fileData.push(data)
+    }
+  })
+  file.on('limit', () => {
+    lastFile.limit = true
+  })
```
mcollina, line 98: is this change relevant?

index.js
```diff
@@ -120,7 +138,10 @@ function fastifyMultipart (fastify, options, done) {
     })
   }
 
-  if (options.attachFieldsToBody === true || options.attachFieldsToBody === 'keyValues') {
+  if (
+    options.attachFieldsToBody === true ||
+    options.attachFieldsToBody === 'keyValues'
+  ) {
```
mcollina, line 144: is this change relevant?

index.js
```diff
@@ -169,25 +191,60 @@ function fastifyMultipart (fastify, options, done) {
...
+  const InvalidJSONFieldError = createError(
+    'FST_INVALID_JSON_FIELD_ERROR',
+    'a request field is not a valid JSON as declared by its Content-Type',
+    406
+  )
+  const FileBufferNotFoundError = createError(
+    'FST_FILE_BUFFER_NOT_FOUND',
+    'the file buffer was not found',
+    500
+  )
```
mcollina, line 238: is this change relevant? Seems just linting

mcollina (changes requested):

index.js
Ceres6: Sorry, linter went crazy, I think they're all solved now

index.js
```diff
@@ -528,6 +535,29 @@ function fastifyMultipart (fastify, options, done) {
     return requestFiles
   }
 
+  async function * filesFromFields (container) {
```
climba03003, line 538: Why async generator but not a generator?
It doesn't really need to be async.

climba03003 (changes requested):

index.js
Ceres6: Already changed

Ceres6:
It seems I can only re-request one review @mcollina @climba03003 @StarpTech, sorry for the inconvenience

climba03003 (approved):

mcollina (approved):
lgtm

PR #431, opened 2023-04-03
Bind original request to `onFile` function `this`

### LINKED ISSUE #427: Option to append `request.id` / other unique ID to `onFile`?
### Prerequisites

- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

On file function gets only `part` and based on filename, you can then save the file to disc etc. This solution is good if the file names are unique, but when 2 users will send the file with the same name, it would become problematic. 
It would be good if second param with `req.id` or other unique ID that would allow to trace the file back to request.
If there is other/easier way - pls let me know. Thanks!

digitalmio:
#### Checklist

Gives access to original request via `this` on `onFile` function. This should be useful to link file handler with original request or create unique IDs based on request id.
Fixes #427

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,7 +157,7 @@ function fastifyMultipart (fastify, options, done) {
         req.body = part.fields
         if (part.file) {
           if (options.onFile) {
-            await options.onFile.bind(req)(part)
+            await options.onFile.call(req, part)
           } else {
             await part.toBuffer()
           }
--- a/index.js
+++ b/index.js
@@ -157,7 +157,7 @@ function fastifyMultipart (fastify, options, done) {
         req.body = part.fields
         if (part.file) {
           if (options.onFile) {
-            await options.onFile(part)
+            await options.onFile.bind(req)(part)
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -157,7 +157,7 @@ function fastifyMultipart (fastify, options, done) {
         req.body = part.fields
         if (part.file) {
           if (options.onFile) {
-            await options.onFile(part)
+            await options.onFile.bind(req)(part)
```
Eomm, line 160:
```suggestion
            await options.onFile.call(req, part)
```

Eomm (approved):
> Not sure if I should change package.json too and bump version to 7.5.1?

Nope

digitalmio:
Any chance to get this approved and merged pls?

jsumners (approved):

Eomm:
when the user sets the arrow function, it should know that the `this` context is the upper scope - as it works in the fastify's handler, decorators and hooks

Issue #173, opened 2020-10-21
multipart does not work with validation and fastify-swagger


xplodeart:
## 🐛 Bug Report

When used with validation and fastify-swagger with the same code from the [example](https://github.com/fastify/fastify-multipart/blob/master/examples/example-validation.js) the file object is empty.

When validation is disabled, the upload works as intended.

## To Reproduce

Register fastify-multipart and fastify-swagger, add a POST route with schema as in the example and try to send a multipart upload with the same name as defined in the schema.

## Expected behavior

Needs to work same as without validation.

## Your Environment

- *node version*: 12
- *fastify version*: 3.7.0
- *os*: Windows

mcollina:
Can you please provide a full example to reproduce?  I'm not sure where the problem is and it will be very helpful. Essentially a server and a client calling it on the same javascript file / or maybe a repo. Essentially something that we can run that clearly show the failure.

xplodeart:
Sure, sorry for making this harder.

Here's a package.json:
`{
  "name": "fastify-bug",
  "version": "0.0.1",
  "description": "",
  "main": "index.js",
  "author": "Max <max.xplodeart@gmail.com>",
  "dependencies": {
    "fastify": "^3.7.0",
    "fastify-multipart": "^3.2.1",
    "fastify-swagger": "^3.4.0",
    "pino-pretty": "^4.3.0"
  }
}
`

Here's a full example of how it won't work:
`const fastify = require( 'fastify' );

const swagger = require( 'fastify-swagger' );

const app = fastify( {
  logger: {
    prettyPrint: true
  },
  ajv: {
    customOptions: {
      jsonPointers: true,
      allErrors: true
    }
  }
} );

app.register( require( 'fastify-multipart' ), {
  attachFieldsToBody: true,
  sharedSchemaId: '#mySharedSchema'
} )

app.register( swagger, {
  routePrefix: '/docs',
  swagger: {
    info: {
      title: 'Sample',
      description: 'Sample',
      version: '0.0.1'
    },
    host: 'localhost:3333',
    schemes: ['http'],
    securityDefinitions: {
      bearerAuth: {
        type: 'apiKey',
        name: 'Authorization',
        in: 'header',
        description: 'Format: "Bearer $TOKEN"'
      }
    },
    consumes: ['application/json','multipart/form-data'],
    produces: ['application/json'],
    tags: [
      {
        name: 'Files',
        description: 'Upload, delete, get signed url'
      }
    ]
  },
[...]

PR #433, opened 2023-04-16
fix: validation example

lord007tn:
#### Checklist

I changed the `validation-example.js` to pass the JSON validation schema.
and add a section in the `readme.md` file so people can have a quick access to make it work
Fixes https://github.com/fastify/fastify-multipart/issues/173


README.md
```diff
@@ -290,6 +290,46 @@ If you try to read from a stream and pipe to a new file, you will obtain an empt
 
 ## JSON Schema body validation
 
+You can use `fastify-multipart` with schema validation, this is a full working example.
```
Eomm, line 293: Could you add a test for this new example?

We can make it consistent in this way

lord007tn:
Still working on It 
further in the project the Ajv validators seems to make fastify fail in some edge cases 
I'm trying to find solutions and I will update it soon 
with tests also

lord007tn:
I added a valid example that worked for me 
but didn't know how to implement unit testing :/

I will give it a shot when I have more time
you could verify it

lord007tn:
I added a working example that i've tried in my work project 
and it works fine 
still lacks unit testing 
i would give it a try when i have more time 
you could verify it

README.md
```diff
@@ -372,6 +374,67 @@ The shared schema, that is added, will look like this:
 }
 ```
 
+### JSON Schema with Swagger
+
+If you want to use `@fastify/multipart` with `@fastify/swagger` and `@fastify/swagger-ui` you must add a new type called `isFiled` and use custom instance of validator compiler [Docs](https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/#validator-compiler).
```
fertolg, line 379:
```suggestion
If you want to use `@fastify/multipart` with `@fastify/swagger` and `@fastify/swagger-ui` you must add a new type called `isFile` and use custom instance of validator compiler [Docs](https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/#validator-compiler).
```

README.md
```diff
@@ -297,6 +297,7 @@ fastify.register(require('@fastify/multipart'), { attachFieldsToBody: 'keyValues
 
 fastify.post('/upload/files', {
   schema: {
+    consumes: ["multipart/form-data"],
```
aj-gameon, line 300:
```suggestion
    consumes: ['multipart/form-data'],
```

README.md
```diff
@@ -331,6 +332,7 @@ fastify.register(require('@fastify/multipart'), opts)
...
```
aj-gameon, line 335:
```suggestion
    consumes: ['multipart/form-data'],
```

README.md
```diff
@@ -463,6 +527,7 @@ fastify.post('/upload/files', {
 })
 ```
 
+
```
aj-gameon, line 530:
```suggestion
```

README.md
```diff
@@ -433,6 +496,7 @@ fastify.register(require('@fastify/multipart'), opts)
 
 fastify.post('/upload/files', {
   schema: {
+    consumes: ["multipart/form-data"],
```
aj-gameon, line 499:
```suggestion
    consumes: ['multipart/form-data'],
```

examples/example-with-swagger.js
```diff
@@ -0,0 +1,60 @@
...
+  keyword: 'isFile',
+  compile: (_schema, parent, _it) => {
+    parent.type = 'file'
+    delete parent.isFile
+    return () => true
+  }
+})
+fastify.setValidatorCompiler(({ schema, method, url, httpPart }) => {
+  return ajv.compile(schema)
+})
```
fertolg, line 31: If you go this route, then any additional schemas will need to be added to the validator instead of the normal `fastify.addSchema` usage, which may not be intuitive. See [the note in the validator compiler docs](https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/#validator-compiler:~:text=Note%3A%20If%20you%20use%20a%20custom%20instance%20of%20any%20validator%20(even%20Ajv)%2C%20you%20have%20to%20add%20schemas%20to%20the%20validator%20instead%20of%20Fastify%2C%20since%20Fastify%27s%20default%20validator%20is%20no%20longer%20used%2C%20and%20Fastify%27s%20addSchema%20method%20has%20no%20idea%20what%20validator%20you%20are%20using.).

Instead of defining an entirely custom validator, would passing the custom keyword as options when initializing fastify allow users to still rely on Fastify's private AJV implementation?

- See [Fastify's AJV factory docs](https://www.fastify.io/docs/latest/Reference/Server/#factory-ajv)
- See the [AJV keywords option](https://ajv.js.org/options.html#keywords)

fertolg:
FWIW, I've been trying to use this instructions in my project for half of today, and I don't think this actually validates properly. If I follow the example I am able to see a file button in Swagger UI, but when attempting to make a request it fails validation with `Error: body must be object`. It's possible I'm missing something else or my specific use case is conflicting with these instructions, are you able to actually make a request using the swagger UI, and the request's body validates properly?

examples/example-with-swagger.js
lord007tn: you made a good point 
as myself I didn't use `addSchema()` but it's valid I will push changes

mcollina (approved):
lgtm

PR #436, opened 2023-04-30
fix!: throw error on file size limit exceeded with addToBody option s…

maxkre96:
…et to true

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,6 +41,10 @@ function attachToBody (options, req, reply, next) {
     return
   }
 
+  const throwFileSizeLimit = typeof options.throwFileSizeLimit === 'boolean'
+    ? options.throwFileSizeLimit
+    : true
+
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

climba03003 (changes requested):
Can you add a test to ensure when multiple files is exist, all of them would be consumed.

I think the current implementation would break on the middle of stream handling and may cause a never consumed multipart happen. Which also means memory leak may happen.

For my own implementation on `busboy`, I just discard all the other files when error comes.
https://github.com/climba03003/fastify-busboy/blob/main/lib/index.ts#L92-L120

gurgunday:
Thanks for taking the time to create this PR

In the latest version of `multipart`, `addToBody` was removed

You can use the newer version, `attachFieldsToBody`, in your applications

PR #443, opened 2023-05-26
Include `ajvFilePlugin` into source and follow OpenAPI convention.

arthurfiorette:
Previously, there was only one example on how to use `@fastify/multipart` together with `@fastify/swagger`. This PR includes the function in the source code to avoid forcing the programmer to always have to open this documentation to see if anything has been changed. And we can be sure that no developer will ever check this again, where future bugfixes will not be fixed for them.
Also, the previous `type: file` has been changed to an `OpenAPI` compatible `type: string, format: binary`.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,7 +384,7 @@ const fastify = require('fastify')({
  // ...
   ajv: {
     // Adds the file plugin to help @fastify/swagger schema generation
-    plugins: [import('@fastify/multipart').ajvFilePlugin]
+    plugins: [require('@fastify/multipart').ajvFilePlugin]
   }
 })
 
--- a/README.md
+++ b/README.md
@@ -379,28 +379,18 @@ The shared schema, that is added, will look like this:
 If you want to use `@fastify/multipart` with `@fastify/swagger` and `@fastify/swagger-ui` you must add a new type called `isFile` and use custom instance of validator compiler [Docs](https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/#validator-compiler).
 
 ```js
-
-const ajvFilePlugin = (ajv, options = {}) => {
- return ajv.addKeyword({
-  keyword: "isFile",
-  compile: (_schema, parent, _it) => {
-   parent.type = "file";
-   delete parent.isFile;
-   return () => true;
-  },
- });
-};
+ 
 const fastify = require('fastify')({
  // ...
   ajv: {
-    // add the new ajv plugin
-    plugins: [/*...*/ ajvFilePlugin]
+    // Adds the file plugin to help @fastify/swagger schema generation
+    plugins: [import('@fastify/multipart').ajvFilePlugin]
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -379,28 +379,18 @@ The shared schema, that is added, will look like this:
...
- });
-};
+ 
 const fastify = require('fastify')({
  // ...
   ajv: {
-    // add the new ajv plugin
-    plugins: [/*...*/ ajvFilePlugin]
+    // Adds the file plugin to help @fastify/swagger schema generation
+    plugins: [import('@fastify/multipart').ajvFilePlugin]
```
climba03003, line 387: `import` return `thenable`, so it doesn't seem correct to me.
For TypeScript, it may works only because all `import` will change to sync `require`.

climba03003 (changes requested):

README.md
arthurfiorette: Nice catch! Changed to `require` to keep consistency with other examples.

arthurfiorette:
@climba03003, sorry for pinging, but any updates on this pr?

felipe-bergamaschi:
@mcollina We need this merge

arthurfiorette:
Hey @climba03003, sorry to ping again, can you give a follow up?

index.d.ts
```diff
@@ -203,6 +203,11 @@ declare namespace fastifyMultipart {
     onFile?: (this: FastifyRequest, part: MultipartFile) => void | Promise<void>;
   }
 
+  /**
+   * Adds a new type `isFile` to help @fastify/swagger generate the correct schema.
+   */
+  export function ajvFilePlugin(ajv: any): void;
```
mcollina, line 209: Can you please add a test for this type? We use tsd.

arthurfiorette: done.

index.js
```diff
@@ -620,6 +620,26 @@ function fastifyMultipart (fastify, options, done) {
...
 
+/**
+ * Adds a new type `isFile` to help @fastify/swagger generate the correct schema.
+ */
+function ajvFilePlugin (ajv) {
+  return ajv.addKeyword({
+    keyword: 'isFile',
+    compile: (_schema, parent) => {
+      // Updates the schema to match the file type
+      parent.type = 'string'
```
climba03003, line 631: Note that changing the actual schema here will have side-effect.
Not sure how it impact but let see what the community feedback.

climba03003 (approved):

index.js
arthurfiorette: What kind of side effect are you talking about? [binary string is what is specified](https://swagger.io/docs/specification/describing-request-body/file-upload/).

climba03003: You are modifying the user input and we don't know how it's behave when he / she re-use such modified schema.
That is what side-effect means.

PR #448, opened 2023-06-21
bugfix: add missing limits.parts type

Curti-s:
When registering the `@fastify/multipart` plugin, esp in TS projects, one cannot be able to provide a limit for parts config.
I believe that, that's not deliberate and that it's also a bug, which this PR addresses.

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -167,6 +167,12 @@ declare namespace fastifyMultipart {
       files?: number;
 
       /**
+       * For multipart forms, the max number of parts (fields + files)
+       * @default 1000
+       */
+      parts?: number;
+
+      /**
        * Max number of header key=>value pairs
        */
       headerPairs?: number;


mcollina (changes requested):
Can you please add a test for this type? We use tsd to test types

gurgunday:
I took a look at the `test-d` file, and I still don't get how we test a config property – just by passing an incorrect type?

index.d.ts
```diff
@@ -166,6 +166,12 @@ declare namespace fastifyMultipart {
        */
       files?: number;
 
+      /**
+       * For multipart forms, the max number of parts (fields + files)
+       * @default 1000
+       */
+      parts?: number;
```
Eomm, line 173: YOu can add the test here: https://github.com/fastify/fastify-multipart/blob/80a01f41c03cc29a074742438b70d95261bb56ff/test/index.test-d.ts#L24

Issue #447, opened 2023-06-12
onFile hook doesn't work on files array field with (attachFieldsToBody: "keyValues")


alifuma92:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.17.0

### Plugin version

7.6.0

### Node.js version

18.15.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.2 (Ventura)

### Description

I want to attach files to my body which also contains other data like strings and booleans.

I used the **attachFieldsToBody: "keyValues"** option, and it seems to work.

**There are however two problems:**

1) the file is attached to the body with the default value buffer.toString(). Here it doesn't appear to convert to a base64 or utf-8 string due to the presence of unknown characters. So it's impossible to restore the buffer and therefore the file from that string. I managed to get around this by using the **onFile hook** which allowed me to return a correct base64 string.

```
onFile: async function (part) {
  const buff = await part.toBuffer();
  const base64_file = buff.toString("base64");
  const type = await fileTypeFromBuffer(buff);
  part.value = {
      file: base64_file,
      type: type,
  };
}
```

2) The problem now is that if I pass more than one file into the same form field, I get returned an array of strings that are not "processed" by the "onFile" hook and are therefore unusable. it's like the onFile hook only works if there is a single file in a specific key of the body, if it's an array it doesn't get executed.

Can anyone come and help me?

### Steps to Reproduce

**1) package.json**

```
{
  "name": "fastifyplayground",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@fastify/multipart": "^7.6.0",
    "fastify": "^4.17.0",
    "file-type": "^18.5.0",
    "mime-types": "^2.1.35"
  },
  "type": "module"
}
```

**2) index.js**

```
import Fastify from "fastify";
import { fileTypeFromBuffer } from "file-type";

// Require the framework and instantiate it
const fastify = Fastify({
    logger: true,
});

// register multipart plugin
fastify.register(import("@fastify/multipart"), {
    attachFieldsToBody: "keyValues",
    limits: {
        fileSize: 1048576, // 1MB
    },
    onFile: async function (part) {
        const buff = await part.toBuffer();
[...]

PR #456, opened 2023-07-12
When options.attachFieldsToBody is set to "keyValues" do not stringify buffer, but return the Buffer

alifuma92:
The details that led to the creation of this pull request are described in this issue.

https://github.com/fastify/fastify-multipart/issues/447

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,12 +172,12 @@ function fastifyMultipart (fastify, options, done) {
           } else if (Array.isArray(field)) {
             body[key] = field.map(item => {
               if (item._buf !== undefined) {
-                return item._buf.toString()
+                return item._buf
               }
               return item.value
             })
           } else if (field._buf !== undefined) {
-            body[key] = field._buf.toString()
+            body[key] = field._buf
           }
         }
         req.body = body


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

gurgunday:
Hey @alifuma92, thank you very much for the PR! I would really appreciate it if we could get this merged, could you please add a unit test when you have the time? ❤️

andersonjoseph:
btw this change was previously discussed [here](https://github.com/fastify/fastify-multipart/pull/348/files/8eaf5f303ea59be087b5dbeb62b40b42894c9c8e#r890340097)

gurgunday:
I couldn’t comment on that discussion, but base64 encoding simply breaks [sharp](https://github.com/lovell/sharp) image processing for me even if I reconvert it back to a buffer

Moreover, not sure why the in-memory buffer is seen as such a bad idea in general **when** you can set a fileSize limit along with a custom error handler to be safe

mcollina:
@gurgunday I don't think this is covering Base64 encoding. Or maybe I misunderstood this PR

gurgunday:
> I don't think this is covering Base64 encoding

Yeah, for some reason I wanted to comment about that [discussion](https://github.com/fastify/fastify-multipart/pull/348/files/8eaf5f303ea59be087b5dbeb62b40b42894c9c8e#r890340097) from here where they discussed using `base64` encoding to stringify the buffer

My issue is that, whatever the type of encoding, once the buffer is stringified, `sharp` won't accept that input even if it's reconverted back, or at least that's what happened to me, which is why I'd like it to just return the buffer instead of doing any conversions

mcollina:
That's what this does, right?

Would you like to take over this work @gurgunday ?

gurgunday:
I'd love to 😁

mcollina:
go for it!

savvykir:
Any updates in here?

PR #461, opened 2023-07-20
Fix file stream consumption when buffer has not loaded yet

### LINKED ISSUE #369: setting onFile callback makes field._buf === null
### Prerequisites

### Fastify version

4.0.1

### Plugin version

7.1.0

### Node.js version

16.14.2

### Operating system

macOS

12.2

### Description

I defined fastify-multipart on a route this way: 
```ts
const onFile = async (part: Multipart) => {
  const url = new Promise<string>((res, rej) => {
    const uploadStream = cloudinary.v2.uploader.upload_stream(
      {},
      (uploadErr, uploadResult) => {
        if (uploadErr || !uploadResult) return rej(uploadErr);

        res(uploadResult.url);
      }
    );
    const result = pipeline(part.file, uploadStream, (err) => {
      if (err) return rej(err);
    });
  });
};


server.register(async (multipartServer) => {
  multipartServer.register(multipart, {
    attachFieldsToBody: "keyValues",
    onFile,
  });

  multipartServer.post("", { schema: ... }, ...);
});
```

when sending a request to the post endpoint i get the following error: `Cannot read properties of null (reading 'toString')`

### Steps to Reproduce

When configuring the fastify-multipart plugin this error is thrown when `attachFieldsToBody` is set to `keyValues` and the `onFile` callback is set.

### Expected Behavior

The reason this error occurs is because when attaching fields to the request body there is only a check whether `field._buffer !== undefined`. `null` is logged when logging the value of `field._buffer`.

andersonjoseph:
#### Checklist

---

Hey! This PR fixes #369

Seems like `_buf` is initialized with `null` and its data is being loaded through `part.toBuffer()`. The condition that checks if the field has data in its buffer compares only against `undefined` instead of `null`, leading to an exception when calling `field._buf.toString()` when `_buf === null`.

Let me know if I'm missing something or if something needs to be changed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,12 +172,12 @@ function fastifyMultipart (fastify, options, done) {
               body[key] = field.value
             } else if (Array.isArray(field)) {
               body[key] = field.map(item => {
-                if (item._buf !== undefined) {
+                if (item._buf) {
                   return item._buf.toString()
                 }
                 return item.value
               })
-            } else if (field._buf !== undefined) {
+            } else if (field._buf) {
               body[key] = field._buf.toString()
             }
           }
--- a/test/multipart-attach-body.test.js
+++ b/test/multipart-attach-body.test.js
@@ -9,6 +9,8 @@ const path = require('path')
 const fs = require('fs')
 const { once } = require('events')
 const { Readable } = require('stream')
+const pump = require('pump')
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #473, opened 2023-08-17
pass field directly for files

### LINKED ISSUE #472: `attachFieldsToBody` makes `request.file()` return undefined.
### Prerequisites

### Fastify version

4.21.0

### Plugin version

7.7.3

### Node.js version

20.3.1

### Operating system

Windows

10.0.19044 Build 19044

### Description

I'm not entirely sure if this is an intentional change, but it's certainly one that I don't personally agree with. As far as I'm aware and have explored, there is no possible way to get the name of the file as `request.body.(field where the file should be)` provides only the buffer and nothing else.

### Steps to Reproduce

```ts
import fastify from "fastify";
import "@fastify/multipart";

const instance = fastify({ logger: true });
instance.register(require("@fastify/multipart"), {
    attachFieldsToBody: "keyValues",
});

instance.route({
    method: "POST",
    url: `/files/`,
    handler: async (request, reply) => {       
        const file = await request.file();
        console.log(file?.filename || "no file"); // returns "no file"
        
        return {
            code: 200,
            message: "ok"
        }
    }
})

instance.listen({ port: 8090 }, (err, address) => {
    if (err) {
        instance.log.error(err);
        process.exit(1);
    }
    instance.log.info(`Server listening on ${address}`);
})
```

### Expected Behavior

Instead of returning `no file`, it should return the data of the sent file without requiring the usage of `request.body.(field where the file should be)`.

gurgunday:
Closes #472

Instead of monkey patching `request.file()` and making it return already consumed file fields, we can just pass them directly to `req.body` when `attachFieldsToBody` is set to `'keyValues'`

Initially, this PR only passed `_buf` and `filename`, but if someone needs to see the file name, they might also want to see the `mimetype` or other properties, so we now just pass the entire field **for files**

#### Checklist


Uzlopak:
Sounds semver major

gurgunday:
Yeah, I pointed it at `next`

gurgunday:
This also lets people use the [registered schema](https://github.com/fastify/fastify-multipart/blob/912d662852d356e8ebc6efe0eb2d7d9c4e1663e4/index.js#L144) to validate file properties, which otherwise is redundant for `attachFieldsToBody: 'keyValues'`

mcollina (approved):
lgtm

gurgunday:
This PR will be superseded

ryanrishi:
@gurgunday was this PR ever superseded by another PR? I'm looking to get both the filename and contents while using `attachFieldsToBody: 'keyValues'`, so this change would still be very helpful.

sourabh-goyal-kasitech:
is this available now?

gurgunday:
Hey, this is still not available

Feel free to send a PR to `next` — you can just pass the file field instead of calling `toBuffer` like in this PR

I'm trying to write a performant `multipart/form-data` parser currently

PR #484, opened 2023-09-28
use `in` to check for prototype violation

gurgunday:
It's 36% faster and provides clearer intent

[Benchmark](https://perf.link/#eyJpZCI6ImwzcnoxZnRxNWxwIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHRlc3QgPSB7fSIsInRlc3RzIjpbeyJuYW1lIjoiRmluZCBpdGVtIDEwMCIsImNvZGUiOiJjb25zdCBhID0gXCJ0b1N0cmluZ1wiXG5pZighKGEgaW4gT2JqZWN0LnByb3RvdHlwZSkpIHRlc3RbYV0gPSAnU0FGRSciLCJydW5zIjpbMzc0NDAwMCw1MjcyMDAwLDQxMTAwMCwyNzcyMDAwLDEzMTQwMDAsMjgyODAwMCwxMDY0NTAwMCwxMjgwNzAwMCwxMzE0MDAwLDUwMDAwLDIzMDQwMDAsMTYzNjAwMCwxMTA0MzAwMCwxOTM1OTAwMCwzMjQ3MDAwLDY3MzMwMDAsMTY3MDgwMDAsMTMxNDAwMCw1ODk5MDAwLDE3NDcwMDAsMTIzMTUwMDAsMTM2NDEwMDAsMTY0NDIwMDAsMTQ5NjAwMCw4MTYxMDAwLDI3MDMwMDAsOTE1MjAwMCwxMzY0NDAwMCwyMTAzMzAwMCwzNDkwMDAwLDEyOTQwMDAwLDEwODUxMDAwLDE3Mzk5MDAwLDI0MDgwMDAsMTA1OTcwMDAsMzQ3ODAwMCwxODQzNDAwMCwzNjk0MDAwLDI3OTIwMDAsODg3OTAwMCw1MTUwMDAsNjE1MDAwLDE2OTM5MDAwLDE0NjUwMDAsNzkzMzAwMCwzMzg0MDAwLDc0NzAwMDAsMTY1MjUwMDAsMzQ0NTAwMCwxMDA0MTAwMCwxNTg2MDAwLDk2NjcwMDAsMTc5MTEwMDAsMTg4NTcwMDAsMTMxNDAwMCw5NjQyMDAwLDEyNzc2MDAwLDE4NDUxMDAwLDEzODIwMDAsMTQ2OTAwMCwxNzA0MDAwLDEzMTQwMDAsMjA0NjkwMDAsMzAyMDAwMCw3OTcyMDAwLDIwNDYzMDAwLDIxMzgyMDAwLDE5NTE2MDAwLDMzNTAwMDAsMTMwMDcwMDAsMjcwNzAwMCw0NTU3MDAwLDI5MDAwLDEyMDU2MDAwLDEzMTQwMDAsMTI3MTkwMDAsMjQ4NjAwMCwzOTAxMDAwLDE5NjM5MDAwLDE3NzkwMDAwLDM2OTEwMDAsNzczMjAwMCwyMTQzODAwMCw0Njc2MDAwLDE5MDg2MDAwLDE5NTkwMDAsMjkwMjAwMCwxOTQ3OTAwMCw0Njk1MDAwLDc1MjgwMDAsMzQxMjAwMCwxOTczMDAwLDE4NTUzMDAwLDIxNzYwMDAsNzIxNjAwMCwyNjQ3MDAwLDczMjYwMDAsMTQ3MzQwMDAsMTk3MzAwMCw4Mzg3MDAwXSwib3BzIjo4MTcwNjEwfSx7Im5hbWUiOiJGaW5kIGl0ZW0gMjAwIiwiY29kZSI6ImNvbnN0IGEgPSBcInRvU3RyaW5nXCJcbmlmKCEoT2JqZWN0LmdldE93blByb3BlcnR5RGVzY3JpcHRvcihPYmplY3QucHJvdG90eXBlLCBhKSkpIHRlc3RbYV0gPSAnU0FGRSciLCJydW5zIjpbMTE4NDAwMCwyNDQ0MDAwLDUxNjcwMDAsMjA5NzAwMCw1MTcwMDAsMTE0MDAwMCw1MzcyMDAwLDEwMzU2MDAwLDEzNTQwMDAsMTE5ODQwMDAsMTE4NDAwMCwzMzAxMDAwLDg4MjEwMDAsMTU3NDEwMDAsMTgwOTAwMCwyNjMzMDAwLDE0NTYzMDAwLDY4MzAwMCwzOTMwMDAwLDI3MjAwMCw2MTg2MDAwLDEwNjc3MDAwLDEzMzE5MDAwLDExODQwMDAsNDYxMTAwMCwyMDU5MDAwLDUxMDkwMDAsMTU4MDYwMDAsMTU4MDYwMDAsMjczNzAwMCw3OTE5MDAwLDg1OTAwMDAsMTM5NTkwMDAsMTkyNTAwMCw2ODUyMDAwLDIxNjIwMDAsMTQ4NzQwMDAsMjQ4NTAwMCwyMTM3MDAwLDU5ODYwMDAsMzg1MDAwLDQ2NzAwMCwxMjgwMjAwMCwxODEzMDAwLDYwOTYwMDAsMTkzMDAwMCw0NTQ2MDAwLDIyNjYwMDAsMTY1NTAwMCw2NDczMDAwLDE2ODAwMCw2NTgxMDAwLDE0NDg3MDAwLDE1NjA0MDAwLDExMDUwMDAsNTY5MzAwMCwxMDIxMzAwMCwxNDQ0MTAwMCwxMjU2NjAwMCw1MjYwMDAsNzU2MDAwLDExNjQwMDAsMTU4MDcwMDAsMTY3ODAwMCwzNjcxMDAwLDE1ODA3MDAwLDE1ODA2MDAwLDE1ODA2MDAwLDIwODIwMDAsNjk3MjAwMCwxMTg0MDAwLDMzOTkwMDAsMzQwMDAsODY4NzAwMCw1MDEwMDAsNjc2NTAwMCwxMjY2MDAwLDExODQwMDAsMTU3ODkwMDAsMTMxNjkwMDAsMjExOTAwMCw5NDM3MDAwLDE1OTA2MDAwLDMxNDYwMDAsMTU4MDYwMDAsMTE4NDAwMCwxODg2MDAwLDE1NzMzMDAwLDI4MzEwMDAsNDA0NTAwMCwxNzEyMDAwLDE1MDkwMDAsMTU1MTkwMDAsMTg2MDAwMCw0Mjk1MDAwLDEwOTUwMDAsMzMwNTAwMCw0MjgxMDAwLDExODQwMDAsNTcxMzAwMF0sIm9wcyI6NjAyODQ1MH1dLCJ1cGRhdGVkIjoiMjAyMy0wOS0yOFQyMjo1NzowNy40NDhaIn0%3D)

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,6 @@ const secureJSON = require('secure-json-parse')
 
 const kMultipart = Symbol('multipart')
 const kMultipartHandler = Symbol('multipartHandler')
-const getDescriptor = Object.getOwnPropertyDescriptor
 
 const PartsLimitError = createError('FST_PARTS_LIMIT', 'reach parts limit', 413)
 const FilesLimitError = createError('FST_FILES_LIMIT', 'reach files limit', 413)
@@ -249,7 +248,7 @@ function fastifyMultipart (fastify, options, done) {
 
     function onField (name, fieldValue, fieldnameTruncated, valueTruncated, encoding, contentType) {
       // don't overwrite prototypes
-      if (getDescriptor(Object.prototype, name)) {
+      if (name in Object.prototype) {
         onError(new PrototypeViolationError())
         return
       }
@@ -295,7 +294,7 @@ function fastifyMultipart (fastify, options, done) {
 
     function onFile (name, file, filename, encoding, mimetype) {
       // don't overwrite prototypes
-      if (getDescriptor(Object.prototype, name)) {
+      if (name in Object.prototype) {
         // ensure that stream is consumed, any error is suppressed
         sendToWormhole(file)
         onError(new PrototypeViolationError())
--- a/test/multipart-security.test.js
+++ b/test/multipart-security.test.js
@@ -104,6 +104,141 @@ test('should not allow __proto__ as field name', function (t) {
   })
 })
 
+test('should not allow toString as field name', function (t) {
+  t.plan(4)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.register(multipart)
+
+  fastify.post('/', async function (req, reply) {
+    t.ok(req.isMultipart())
+
+    try {
+      await req.file()
+      reply.code(200).send()
+    } catch (error) {
+      t.ok(error instanceof fastify.multipartErrors.PrototypeViolationError)
+      reply.code(500).send()
+    }
+  })
+
+  fastify.listen({ port: 0 }, async function () {
+    // request
+    const form = new FormData()
+    const opts = {
+      protocol: 'http:',
+      hostname: 'localhost',
+      port: fastify.server.address().port,
+      path: '/',
+      headers: form.getHeaders(),
+      method: 'POST'
+    }
+
+    const req = http.request(opts, (res) => {
+      t.equal(res.statusCode, 500)
+      res.resume()
+      res.on('end', () => {
+        t.pass('res ended successfully')
+      })
+    })
+    form.append('toString', 'world')
+
+    form.pipe(req)
+  })
+})
+
+test('should not allow hasOwnProperty as field name', function (t) {
+  t.plan(4)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.register(multipart)
+
+  fastify.post('/', async function (req, reply) {
+    t.ok(req.isMultipart())
+
+    try {
+      await req.file()
+      reply.code(200).send()
+    } catch (error) {
+      t.ok(error instanceof fastify.multipartErrors.PrototypeViolationError)
+      reply.code(500).send()
+    }
+  })
+
+  fastify.listen({ port: 0 }, async function () {
+    // request
+    const form = new FormData()
+    const opts = {
+      protocol: 'http:',
+      hostname: 'localhost',
+      port: fastify.server.address().port,
+      path: '/',
+      headers: form.getHeaders(),
+      method: 'POST'
+    }
+
+    const req = http.request(opts, (res) => {
+      t.equal(res.statusCode, 500)
+      res.resume()
+      res.on('end', () => {
+        t.pass('res ended successfully')
+      })
+    })
+    form.append('hasOwnProperty', 'world')
+
+    form.pipe(req)
+  })
+})
+
+test('should not allow propertyIsEnumerable as field name', function (t) {
+  t.plan(4)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.register(multipart)
+
+  fastify.post('/', async function (req, reply) {
+    t.ok(req.isMultipart())
+
+    try {
+      await req.file()
+      reply.code(200).send()
+    } catch (error) {
+      t.ok(error instanceof fastify.multipartErrors.PrototypeViolationError)
+      reply.code(500).send()
+    }
+  })
+
+  fastify.listen({ port: 0 }, async function () {
+    // request
+    const form = new FormData()
+    const opts = {
+      protocol: 'http:',
+      hostname: 'localhost',
+      port: fastify.server.address().port,
+      path: '/',
+      headers: form.getHeaders(),
+      method: 'POST'
+    }
+
+    const req = http.request(opts, (res) => {
+      t.equal(res.statusCode, 500)
+      res.resume()
+      res.on('end', () => {
+        t.pass('res ended successfully')
+      })
+    })
+    form.append('propertyIsEnumerable', 'world')
+
+    form.pipe(req)
+  })
+})
+
 test('should use default for fileSize', async function (t) {
   t.plan(4)
 


Uzlopak:
Well. hmm. why not just do `const body = { __proto__ }`

or more performant:

```
const Body = function NullObject () { }
Body.prototype = Object.create(null)


const body = new Body()
```

gurgunday:
I really prefer the second one or simply doing `body = Object.create(null)` and removing these prototype checks, but I'm afraid it'd be a breaking change 😁

WDYT?

gurgunday:
Or maybe this is to detect malicious user behavior and respond accordingly since it throws a distinct error

Uzlopak:
Object.create(null) is super slow.

imho this should be discussed. same as we did with https://github.com/delvedor/find-my-way/pull/333

Well. hmm. i should write that benchmark i promised few weeks ago for find-my-way.

gurgunday:
Wow, the performance difference is huge 😄

Uzlopak:
@mcollina 
@climba03003 
@Eomm 

How about making the body null prototype?

climba03003 (changes requested):
`in` will look through the prototype chain, so a no in here

gurgunday:
Yeah, but the prototype chain of `Object.prototype`, which isn't further extended — `Object.prototype.__proto__ === null`

How is it different than calling getOwnDescriptor on `Object.prototype`?

In any case, I think it's best if we go with NullObject and remove these checks — the body is simply representing a Map in this context and a user should only be accessing its keys

gurgunday:
> Then you can allow even "bad" names.

@Uzlopak not sure anymore about that... I don't think it should be possible with multipart but if someone sneaks in a custom__proto__ object then you certainly have prototype poisoning

So the checks must stay, and if they stay, what's the benefit of having a null prototype?

```js
function NullObject () {}
NullObject.prototype = Object.create(null)
// {}
a = new NullObject()
// NullObject {}
a
// NullObject {}[[Prototype]]: ObjectNo properties
a['__proto__'] = {shh: 5}
// {shh: 5}
b = Object.assign({}, a)
// {}
b.shh
// 5
```

Eomm:
> How about making the body null prototype?

I think was already planned for fastify v5. isn't it?

Uzlopak:
I think this was referring for parameters object. idk.

gurgunday:
I think just to align the multipart parser with the default JSON parser, we should make `req.body`'s prototype a null-prototype object in a future PR

gurgunday:
Can I persuade someone to approve? 😁

To reiterate, `Object.prototype` **cannot** have a non-null `__proto__` since it's the end of the prototype chain, so using `in` here is equivalent to `getOwnPropertyDescriptor` - but faster

This should be a free performance improvement 🤠

Uzlopak:
Well. Do we had before this PR unit tests covering prototype pollution checks?

gurgunday:
We have checks for __proto__ pollution:

https://github.com/fastify/fastify-multipart/blob/109a1165bbba8571f178dbee9b30955a2556f61f/test/multipart-security.test.js#L16-L105

Let me add a few more

Uzlopak:
Ah, i think there is also another issuer with using `in`. As @climba03003 wrote, it means we go up the prototype chain. 

If we dont use a null object this means, that we use Object. Object has e.g.`'hasOwnProperty`, which is a set to enumerable false. So if you use `Object.keys(body)`, you dont see it. So if multipart has a key named `hasOwnProperty` it is no issue, because we can overwrite it, as `Object.prototype.hasOwnProperty.call(body, 'hasOwnProperty')` returns `false`. And `Object.keys(body)` will contain `hasOwnProperty`.

If this PR lands, it will make it impossible to use keys like `hasOwnProperty`. 

That is my assessment of @climba03003 remark. I am not at my PC to check it. 

Can you check it please?

gurgunday:
No, it is much more simpler than that

`master` (OLD) already blocks these, we are already checking if a name is IN `Object.property` by applying `getOwnPropertyDescriptor` **directly** on `Object.prototype` with the field name

For the prototype chain, ECMAScript defines that `Object.property` is the **last** step of the prototype chain, its value is `null` and cannot go further – as an example, try modifying `Object.prototype.__proto__`, it will result in an error

So instead of calling `getOwnPropertyDescriptor`, we are simply using the `in` keyword, which, yes looks at the prototype chain, but it's already the last step! It's completely the same in terms of semantics

Uzlopak:
Ah, i see.

@climba03003 
Can you please revise your review based on the comments of @gurgunday ?

Uzlopak (approved):
Approving, but please dont merge before we did not convinced @climba03003

climba03003 (approved):
Yes, after a few test `Object.prototype` is allowed to modify.

PR #502, opened 2023-12-27
Partly closes #496: add missing type

### LINKED ISSUE #496: Incorrect behavior of per-request limits
### Prerequisites

### Fastify version

4.24.3

### Plugin version

8.0.0

### Node.js version

21.2.0

### Operating system

Windows

W11. 23H2 (22631.2715)

### Description

Clear and concise:
1. Inaccuracy in documentation (possibly);
2. Inability to specify limits for each route separately.

I also want to clarify that I am using TypeScript.
TypeScript version: 5.3.2.

### Steps to Reproduce

1. From plugin types: 
image
In `BusboyConfig` interface simply no `throwFileSizeLimit`.

2. Set limits for `request.files()`
image
Set `throwFileSizeLimit: true` globally
image

Send larger files and in larger quantities. As a result, we do NOT get exception.
However, If you specify the same parameters globally or for the `saveRequestFiles` function, the limits will work correctly.

### Expected Behavior

1. The first thing I would like to say is that the instructions in the documentation are incorrect (or am I wrong?).
Example from docs:
image
However, I cannot specify such a parameter due to differences in types. `files()` allow `BusboyConfig` interface, in which there is no `throwFileSizeLimit` parameter. It is only available in the `FastifyMultipartBaseOptions` interface.
image

2. What I'm trying to do: set basic global limits when registering a plugin and specify the limits of each route separately, for part of them, as an overwrite of the global ones.
Global:
```ts
 const fastifyMultipartOptions: FastifyMultipartOptions = {
    limits: {
      fieldNameSize: 100,
      fieldSize: 100,
      fields: 5,
      fileSize: 1,
      headerPairs: 2000,
      files: 5,
    },
    throwFileSizeLimit: true,
  };
  await app.register(fastifyMultipart, fastifyMultipartOptions);
```
For route (I can't set `throwFileSizeLimit` because of the problem described earlier):
`await request.files({ limits: { files: 1, fileSize: 1 } });`

The limits specified globally work correctly, but the specified limits for a request are completely ignored.
I tried this way: use `saveRequestFiles(opts)`, where setting limits works correctly, however, I cannot get `toBuffer()` from the `savedRequestFiles` array, which makes this function not relevant for me.

hash0000000:
Fix type missing. More info in issue #496.

#### Checklist

### DIFF
--- a/test/multipart-update-options-type.test.js
+++ b/test/multipart-update-options-type.test.js
@@ -8,7 +8,7 @@ const http = require('node:http')
 const EventEmitter = require('node:events')
 const { once } = EventEmitter
 
-test('Should throw RequestFileTooLargeError when throwFileSizeLimit: false for file())', async function (t) {
+test('Should throw RequestFileTooLargeError when throwFileSizeLimit: true for file())', async function (t) {
   t.plan(3)
 
   const fastify = Fastify()
@@ -112,7 +112,7 @@ test('Should NOT throw RequestFileTooLargeError when throwFileSizeLimit: false f
   }
 })
 
-test('Should throw RequestFileTooLargeError when throwFileSizeLimit: false for files())', async function (t) {
+test('Should throw RequestFileTooLargeError when throwFileSizeLimit: true for files())', async function (t) {
   t.plan(3)
 
   const fastify = Fastify()
--- a/test/multipart-update-options-type.test.js
+++ b/test/multipart-update-options-type.test.js
@@ -1,0 +1,221 @@
+'use strict'
+
+const test = require('tap').test
+const FormData = require('form-data')
+const Fastify = require('fastify')
+const multipart = require('..')
+const http = require('node:http')
+const EventEmitter = require('node:events')
+const { once } = EventEmitter
+
+test('Should throw RequestFileTooLargeError when throwFileSizeLimit: false for file())', async function (t) {
+  t.plan(3)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.register(multipart)
+
+  fastify.post('/', async function (req, reply) {
+    t.ok(req.isMultipart())
+
+    try {
+      const file = await req.file({ limits: { fileSize: 1 }, throwFileSizeLimit: true })
+      await file.toBuffer()
+      t.fail('should throw')
+      reply.code(200).send()
+    } catch (error) {
+      t.ok(error instanceof fastify.multipartErrors.RequestFileTooLargeError)
+      reply.code(500).send()
+    }
+  })
+
+  await fastify.listen({ port: 0 })
+  // request
+  const form = new FormData()
+  const opts = {
+    protocol: 'http:',
+    hostname: '127.0.0.1',
+    port: fastify.server.address().port,
+    path: '/',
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for our types.

mcollina:
The type test is https://github.com/fastify/fastify-multipart/blob/master/types/index.test-d.ts

hash0000:
Does the `files()` and `file()` support all options from the `FastifyMultipartBaseOptions` interface (except `throwFileSizeLimit` and `limits`, of course)?

mcollina:
I think so, but verify with the code.

hash0000:
I'm not sure, but is this the test you meant?

Uzlopak:
Lets run the tests. Your holistic approach to also test the same thing in javascript is apprecaited.

types/index.test-d.ts
```diff
@@ -85,7 +95,18 @@ const runServer = async () => {
...
+      throwFileSizeLimit: true,
+      sharedSchemaId: 'schemaId',
+      isPartAFile: (fieldName, contentType, fileName) => {
+        expectType<string | undefined>(fieldName)
+        expectType<string | undefined>(contentType)
+        expectType<string | undefined>(fileName)
+        return true
+      }
+    }
+    const parts = req.files(options)
```
Uzlopak, line 109: hmm. You are type casting options. 

```suggestion
    const parts = req.files({
      limits: { fileSize: 1000, parts: 500 },
      throwFileSizeLimit: true,
      sharedSchemaId: 'schemaId',
      isPartAFile: (fieldName, contentType, fileName) => {
        expectType<string | undefined>(fieldName)
        expectType<string | undefined>(contentType)
        expectType<string | undefined>(fileName)
        return true
      }
    })
```

mcollina (approved):
lgtm

test/multipart-update-options-type.test.js
```diff
@@ -0,0 +1,221 @@
...
+
+const test = require('tap').test
+const FormData = require('form-data')
+const Fastify = require('fastify')
+const multipart = require('..')
+const http = require('node:http')
+const EventEmitter = require('node:events')
+const { once } = EventEmitter
+
+test('Should throw RequestFileTooLargeError when throwFileSizeLimit: false for file())', async function (t) {
```
gurgunday, line 11:
```suggestion
test('Should throw RequestFileTooLargeError when throwFileSizeLimit: true for file())', async function (t) {

```

test/multipart-update-options-type.test.js
```diff
@@ -0,0 +1,221 @@
...
+    const [res] = await once(req, 'response')
+    t.equal(res.statusCode, 200)
+    res.resume()
+    await once(res, 'end')
+  } catch (error) {
+    t.error(error, 'request')
+  }
+})
+
+test('Should throw RequestFileTooLargeError when throwFileSizeLimit: false for files())', async function (t) {
```
gurgunday, line 115:
```suggestion
test('Should throw RequestFileTooLargeError when throwFileSizeLimit: true for files())', async function (t) {
```

mcollina (approved):
lgtm

PR #506, opened 2024-01-24
use @fastify/busboy v2.1.0

gurgunday:
Motivation: busboy@2 is more strict

For now it's not ideal (string comparison) but it works, maybe we can return a standard error message or error code from busboy, like `400`


mcollina (approved):
lgtm

PR #507, opened 2024-01-31
fix: make sure the handler resolves in all cases

### LINKED ISSUE #497: Request hangs if an error is thrown while processing a file
### Prerequisites

### Fastify version

4.24.3

### Plugin version

8.0.0

### Node.js version

18.x

### Operating system

Linux

20.04

### Description

The `for..await of` promise doesn't get fulfilled (or rejected) if an error is throw when processing a single file. For example, in the following code, the line "done processing all files" doesn't get logged.

```js
const files = await request.files();

for await (const { file, filename } of files) {
  console.log("start processing file", filename);

  try {
    const storage = new Writable({
      write(chunk, encoding, callback) {
        // trigger error:
        callback(new Error("write error"));
      },
    });

    await pipeline(file, storage);
  } catch (error) {
    console.log("caught error while processing file", filename, error);
    // note that the error isn't rethrown here, so this error is now handled
    // and processing should proceed
  }

  console.log("done processing file", filename);
}

console.log("done processing all files");

return reply.send(files);
```

Also, if the `try..catch` is removed, everything freezes after "start processing file".

On a separate/related note, is it possible to use `@fastify/multipart` but with the stream-based api instead of the generators? 

### Steps to Reproduce

Repro: https://github.com/joelmukuthu/fastify-multipart-hanging-bug.

### Expected Behavior

That the request would not hang. If the error is not handled properly (e.g. if there's no `try..catch` in the example above), I'd expect that the request fails with an error.

gurgunday:
partially fixes #497

The request still freezes without the `try catch` and I'm not sure how we could remedy that


climba03003 (commented):
Seems like it was a wrong choice to use `while` for generator.
I knew it will cause problem somehow, that's why I prefer to use [`AsyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncIterator)

Something similar can be done by,
https://github.com/kaka-ng/fastify-plugins/blob/main/packages/multipart/lib/adapter/busboy.ts#L116-L156

gurgunday:
I don't get an error on node 16, 18, 20, or 21

What is up with 14?..

mcollina (approved):
lgtm

gurgunday:
Alright, I'll simplify (I tried to go overboard so that it resolves in node 14 but it just won't catch the error...)

Should we merge like this or should I merge to next?

Uzlopak:
@gurgunday 

Can you shortly describe why it fails on 14?

gurgunday:
I can't for the moment 🤣

Essentially, the test I added doesn't pass on node 14, and it needs investigation

gurgunday:
Note that it's not because of what this PR does, it is because the PR doesn't fix the issue on node 14

gurgunday:
Seems like on node 14 request.raw.on('error') is not firing... anyone has a guess on why that might be? Was it added recently?

mcollina:
I think the destroy logic in v14 is _really different_.

mcollina (approved):
lgtm

climba03003 (approved):

Uzlopak (approved):
RSLGTM

gurgunday:
Will cut a patch

mcollina:
a patch? This is in the next branch? Is that ready to ship?

gurgunday:
> a patch? This is in the next branch? Is that ready to ship?

Yeah, forgot for a second

It's not ready IMO, we can wait till v5 since maybe I'll be able to do the major rewrite... unless someone needs this patch ASAP

joelmukuthu:
Thanks @gurgunday (and reviewers) for working on this fix. I understand that there are some issues with getting this released due to node 14, but is there any chance to get this released in a bug-fix release? Or do you have a rough timeline for the v5 release?

gurgunday:
@mcollina, should I add a skip to the test and backport it to main? Or would you prefer releasing next?

PR #522, opened 2024-04-30
Add FormData decorator to request

mcollina:
Adds a tiny helper to create a `FormData` object for compatibility with Web Standards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ fastify.post('/upload/files', async function (req, reply) {
   ) // Request body in key-value pairs, like req.body in Express (Node 12+)
 
   // On Node 18+
-  const formData = await req.formdata()
+  const formData = await req.formData()
   console.log(formData)
 })
 ```
--- a/README.md
+++ b/README.md
@@ -235,6 +235,10 @@ fastify.post('/upload/files', async function (req, reply) {
   const body = Object.fromEntries(
     Object.keys(req.body).map((key) => [key, req.body[key].value])
   ) // Request body in key-value pairs, like req.body in Express (Node 12+)
+
+  // On Node 18+
+  const formData = await req.formdata()
+  console.log(formData)
 })
 ```
 
@@ -518,6 +522,7 @@ fastify.post('/upload/files', async function (req, reply) {
 This project is kindly sponsored by:
 - [nearForm](https://nearform.com)
 - [LetzDoIt](https://www.letzdoitapp.com/)
+- [platformatic](https://platformatic.dev)
 
 ## License
 
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ const PrototypeViolationError = createError('FST_PROTO_VIOLATION', 'prototype pr
 const InvalidMultipartContentTypeError = createError('FST_INVALID_MULTIPART_CONTENT_TYPE', 'the request is not multipart', 406)
 const InvalidJSONFieldError = createError('FST_INVALID_JSON_FIELD_ERROR', 'a request field is not a valid JSON as declared by its Content-Type', 406)
 const FileBufferNotFoundError = createError('FST_FILE_BUFFER_NOT_FOUND', 'the file buffer was not found', 500)
+const NoFormData = createError('FST_NO_FORM_DATA', 'FormData is not available', 500)
 
 function setMultipart (req, payload, done) {
   req[kMultipart] = true
@@ -157,6 +158,12 @@ function fastifyMultipart (fastify, options, done) {
         return formData
       })
     }
+  } 
+
+  if (!fastify.hasRequestDecorator('formData')) {
+    fastify.decorateRequest('formData', async function () {
+      throw new NoFormData()
+    })
   }
 
   const defaultThrowFileSizeLimit = typeof options.throwFileSizeLimit === 'boolean'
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -235,6 +235,10 @@ fastify.post('/upload/files', async function (req, reply) {
   const body = Object.fromEntries(
     Object.keys(req.body).map((key) => [key, req.body[key].value])
   ) // Request body in key-value pairs, like req.body in Express (Node 12+)
+
+  // On Node 18+
+  const formData = await req.formdata()
```
gurgunday, line 240: Awesome addition, literally what I wanted to do in the rewrite

Why `req.formdata` instead of `req.formData`?

index.js
```diff
@@ -123,6 +123,40 @@ function fastifyMultipart (fastify, options, done) {
...
+            for (const item of value) {
+              await append(key, item)
+            }
+          } else {
+            await append(key, value)
+          }
+        }
+
+        async function append (key, entry) {
+          if (entry.type === 'file') {
```
gurgunday, line 145: This won't work with `attachFieldsToBody: 'keyValues'`

```suggestion
          if (entry.type === 'file' || (attachFieldsToBody === 'keyValues' && Buffer.isBuffer(entry))) {
```

`'keyValues'` comparison could be dropped for performance, it should be fine, or you could just check if it's type is string here

But it becomes harder to understand

index.js
```diff
@@ -123,6 +123,40 @@ function fastifyMultipart (fastify, options, done) {
...
+            // TODO use File constructor with fs.openAsBlob()
+            // if attachFieldsToBody is not set
+            // https://nodejs.org/api/fs.html#fsopenasblobpath-options
+            formData.append(key, new Blob([await entry.toBuffer()], {
+              type: entry.mimetype
+            }), entry.filename)
```
gurgunday, line 151: Actually yeah this part won't work at all, maybe we can decorate `formData` if only keyvalues is not set

mcollina:
@gurgunday do you know why linting is failing, it's passing locally.

gurgunday:
Seems like a dependency conflict

`"eslint-plugin-n"` was updated to v17:
https://github.com/fastify/fastify-multipart/blob/7d6e1247dc34e3c9e121fbf38585f36611e2a3ce/package.json#L28

And standard requires 16 normally...

Running the following will solve it:

```bash
npm i eslint-plugin-n@16
```

gurgunday:
I actually can't see why they were installed separately in this repo

gurgunday:
Pulling from master should fix it

index.js
```diff
@@ -123,6 +123,40 @@ function fastifyMultipart (fastify, options, done) {
         req.body = body
       }
     })
+
+    // The following is not available on old Node.js versions
+    // so we must skip it in the test coverage
+    /* istanbul ignore next */
+    if (globalThis.FormData) {
+      fastify.decorateRequest('formData', async function () {
```
kibertoad, line 131: probably TS types need to be updated too?

index.js
```diff
@@ -123,6 +123,40 @@ function fastifyMultipart (fastify, options, done) {
...
+          if (Array.isArray(value)) {
+            for (const item of value) {
+              await append(key, item)
+            }
+          } else {
+            await append(key, value)
+          }
+        }
+
+        async function append (key, entry) {
```
kibertoad, line 144: there are no awaits, does this need to be async?

gurgunday: Yeah I thought the same for a second: https://github.com/fastify/fastify-multipart/pull/522/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R149

kibertoad: thanks :D

mcollina:
@kibertoad fixed

gurgunday (commented):
Last change broke it for node 14, 16

index.js
```diff
@@ -123,6 +124,46 @@ function fastifyMultipart (fastify, options, done) {
...
+        }
+
+        return formData
+      })
+    }
+  } 
+
+  if (!fastify.hasRequestDecorator('formData')) {
+    fastify.decorateRequest('formData', async function () {
+      throw new NoFormData()
```
gurgunday, line 165: This line is not covered in the tests, and lint is failing

index.js
```diff
@@ -123,6 +124,47 @@ function fastifyMultipart (fastify, options, done) {
         req.body = body
       }
     })
+
+    // The following is not available on old Node.js versions
+    // so we must skip it in the test coverage
+    /* istanbul ignore next */
+    if (globalThis.FormData) {
+      fastify.decorateRequest('formData', async function () {
```
gurgunday, line 132: Would you consider this a breaking change for the same reason as in https://github.com/fastify/fastify-env/pull/183#discussion_r1590240512?

That `formData` might already be decorated since it's a common name?

mcollina: Done

kibertoad (approved):

gurgunday (approved):
lgtm

gurgunday (approved):

gurgunday:
@Eomm @mcollina did we add a new rule to this plugin?

Can't release `v8.3.0`

Screenshot 2024-05-11 at 00 31 20

ImUrX:
when is a new release expected?

PR #533, opened 2024-07-09
Merge `next` into `master`

jsumners:
As titled.


index.js
```diff
@@ -7,12 +7,11 @@ const { createWriteStream } = require('node:fs')
 const { unlink } = require('node:fs/promises')
 const path = require('node:path')
 const { generateId } = require('./lib/generateId')
-const util = require('node:util')
 const createError = require('@fastify/error')
-const sendToWormhole = require('stream-wormhole')
+const streamToNull = require('./lib/stream-consumer')
```
jsumners, line 11: I had to rip out this dependency because it was hanging up the `big.test.js` test (I think it was never consuming the stream). When I got done fixing the coverage requirement to get `npm test` to return `$? === 0`, I thought I already had a commit with the replacement and did a `git commit --amend -n` and couldn't figure out how to quit Vim without the commit going through. So it got amended into the rebase commit 🤷‍♂️

index.js
```diff
@@ -349,7 +349,7 @@ function fastifyMultipart (fastify, options, done) {
       // don't overwrite prototypes
       if (name in Object.prototype) {
         // ensure that stream is consumed, any error is suppressed
-        sendToWormhole(file)
+        streamToNull(file)
```
gurgunday, line 352: So this is a Promise (that can reject) and there isn't a `.catch` on it

It can bring everything down

The original package does not reject, or you can add a .catch here

```suggestion
        streamToNull(file).catch(() => {})
```

gurgunday (approved):

PR #548, opened 2024-10-08
Support firebase

ThomasKientz:
Currently we can't use this plugin with [serverless](https://fastify.dev/docs/latest/Guides/Serverless/) GCP (or Firebase).

As explained here https://github.com/fastify/fastify/issues/946#issuecomment-766319521 and [in the doc for firebase](https://fastify.dev/docs/latest/Guides/Serverless/#add-custom-contenttypeparser-to-fastify-instance-and-define-endpoints), we need to use `req.rawBody` in order to parse files from multipart.

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,6 +30,7 @@ const NoFormData = createError('FST_NO_FORM_DATA', 'FormData is not available', 
 function setMultipart (req, payload, done) {
   req[kMultipart] = true
   req.rawBody = payload.rawBody
+  fastify.decorateRequest('rawBody', req.rawBody)
   done()
 }
 
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ const NoFormData = createError('FST_NO_FORM_DATA', 'FormData is not available', 
 
 function setMultipart (req, payload, done) {
   req[kMultipart] = true
+  req.rawBody = payload.rawBody
   done()
 }
 
@@ -297,7 +298,11 @@ function fastifyMultipart (fastify, options, done) {
       process.nextTick(() => cleanup(err))
     })
 
-    request.pipe(bb)
+    if (request.rawBody) {
+      bb.end(request.rawBody)
+    } else {
+      request.pipe(bb)
+    }
 
     function onField (name, fieldValue, fieldnameTruncated, valueTruncated, encoding, contentType) {
       // don't overwrite prototypes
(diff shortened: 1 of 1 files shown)


Uzlopak:
the tests are not happy

gurgunday (changes requested):
Decorate that new assignment to req please

index.js
```diff
@@ -30,6 +30,7 @@ const NoFormData = createError('FST_NO_FORM_DATA', 'FormData is not available',
 function setMultipart (req, payload, done) {
   req[kMultipart] = true
   req.rawBody = payload.rawBody
+  fastify.decorateRequest('rawBody', req.rawBody)
```
Uzlopak, line 33: you decorate the request when you register the plugin. then yo can assign rawbody to the fastify request.

ThomasKientz:
> the tests are not happy

 @Uzlopak don't really understand why it's failing, looks good but return exit code 1

Uzlopak:
The test coverage is Not 100%

climba03003 (commented):
Where is `payload.rawBody` comes from?
I don't like the current idea to support one serverless environment and adding extra branching.

Instead one should provide a `plugin` that dedicated for the serverless environment and add all the necessary patching to match how `Node.js` works.

Uzlopak:
Well the problem is, that gcf is a glorified express router. So they preparse the body so that you dont have to. You get the body stream in rawBody. So maybe we need, like you suggest, a plugin which basically removes the default content type parsers and just assigns rawbody to body?!

climba03003:
What's comes my mind is something like?

```js
function gcp(fastify) {
  // add pass through content-type parser
  fastify.addContentTypeParser('application/json', {}, (req, body, done) => {
    done(null, body.body);
  });
  
  // add stream transform for multipart when neccessary
  if (fastify.hasContentTypeParser('multipart/form-data')) {
    fastify.addHook('preParsing', function(request, reply, payload, done) {
      if(request.headers['content-type'].startsWith('multipart/form-data')) {
        // we override the `.pipe` method and return rawBody as new payload
        const pipe = request.pipe
        request.pipe = function(stream) {
          Readable.from([rawBody]).pipe(stream)
        }
        done(null, payload.rawBody)
      } else {
        done(null, payload)
      }
    })
  }
}
```

ThomasKientz:
Should I pursue the current implementation of this pull request and add test coverage? @Uzlopak

PR #573, opened 2025-03-16
Migrate from tap to Node Test Runner

gotamm:
Hi,

The goal of this PR is migrate the tests from Tap to Node test runner, following what's been said in the issue https://github.com/fastify/fastify/issues/5555

#### Checklist

### DIFF
--- a/test/multipart-security.test.js
+++ b/test/multipart-security.test.js
@@ -263,7 +263,7 @@ test('should use default for fileSize', async function (t) {
     try {
       await part.toBuffer()
       reply.send('not ok')
-      t.assert.ok('it should throw')
+      t.assert.fail('it should throw')
     } catch (error) {
       t.assert.ok(error)
       reply.send(error)
@@ -311,7 +311,7 @@ test('should use default for parts - 1000', function (t, done) {
     try {
       // eslint-disable-next-lint no-empty
       for await (const _ of req.parts()) { console.assert(_) }
-      t.assert.ok('should throw on 1001')
+      t.assert.fail('should throw on 1001')
       reply.code(200).send()
     } catch (error) {
       t.assert.ok(error instanceof fastify.multipartErrors.PartsLimitError)
--- a/test/multipart-security.test.js
+++ b/test/multipart-security.test.js
@@ -297,7 +297,7 @@ test('should use default for fileSize', async function (t) {
     res.resume()
     await once(res, 'end')
   } catch (error) {
-    t.assert.ok(error, 'request')
+    t.assert.ifError(error)
   }
 })
 
--- a/test/multipart-security.test.js
+++ b/test/multipart-security.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const test = require('tap').test
+const test = require('node:test')
 const FormData = require('form-data')
 const Fastify = require('fastify')
 const multipart = require('..')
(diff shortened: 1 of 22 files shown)


Fdawgs:
Thanks for the PR! `.taprc` also needs removing as part of this (but kept in `.gitignore`).

Eomm:
Could you run `npm run lint:fix`? it should fix the CI

matteo-gobbo:
The issue here is using const `tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), ''))` instead of the previous `t.testDir()` that was implemented directly inside tap.
Do you have any suggestions about this? Perhaps there are other projects where a similar approach has been implemented that I could use as a reference?"

Thank you in advance for your answer.

Eomm:
> The issue here is using `const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), ''))` instead of the previous `t.testDir()` that was implemented directly inside tap. Do you have any suggestions about this? Perhaps there are other projects where a similar approach has been implemented that I could use as a reference?"
> 
> Thank you in advance for your answer.

If i check the [`testdir` implementation](https://github.com/tapjs/tapjs/blob/532287f65d814c7cb7872bf84b3d86b17d900e43/src/fixture/src/index.ts#L92) I think it was using a path in the project's folder (such as `.tap/fixtures`).
So I think we may not create the `ostmp` by a random folder in a similar way (that we must add to the gitignore)

ilteoood:
@matteo-gobbo are you planning to complete this?

matteo-gobbo:
> What suggested by @Eomm is a good starting point. At the moment there's no other project that has been migrated that used testdir. But if you check your inner implementation as suggested, it doesn't anything special. You just need to point to the right path.
> 
> Let me know if you need additional help, I'll try to provide additional support if you can't figure it out.

I’m actually out for the next couple of days, but I’ll give it a try as soon as I’m back. 
Thanks for your help!

gurgunday (changes requested):
lgtm other than these

gurgunday (approved):
lgtm

test/multipart-small-stream.test.js
```diff
@@ -51,24 +50,24 @@ test('should throw fileSize limitation error on small payload', { skip: true },
 
   try {
     const [res] = await once(req, 'response')
-    t.equal(res.statusCode, 413)
+    t.assert.strictEqual(res.statusCode, 413)
     res.resume()
     await once(res, 'end')
   } catch (error) {
-    t.error(error, 'request')
+    t.assert.ok(error, 'request')
```
is2ei, line 57: @matteo-gobbo 

`t.error` is not the same as `t.assert.ok`.

https://tapjs.github.io/tapjs/classes/_tapjs_test.index.Test.html#error
> Assert that an error object is not provided.

https://nodejs.org/api/assert.html#assertokvalue-message
> Tests if value is truthy.

is2ei (changes requested):

test/multipart-security.test.js
```diff
@@ -288,28 +293,28 @@ test('should use default for fileSize', async function (t) {
...
+  t.after(() => fastify.close())
 
   fastify.register(multipart)
 
   fastify.post('/', async function (req, reply) {
     try {
       // eslint-disable-next-lint no-empty
       for await (const _ of req.parts()) { console.assert(_) }
-      t.fail('should throw on 1001')
+      t.assert.ok('should throw on 1001')
```
is2ei, line 314: @matteo-gobbo 

Just to clarify, `t.assert.ok` isn’t the same as `t.fail`.  
`assert` also provides a `fail` method, which might be helpful here:  

https://nodejs.org/api/assert.html#assertfailmessage

is2ei (changes requested):

is2ei (approved):
LGTM!

gurgunday (approved):

mcollina (approved):
lgtm

gurgunday (approved):
lgtm

gurgunday:
Can you rebase?

dancastillo (approved):

p-kuen:
In this PR the `c8` package was added to the dependencies instead of just dev dependencies. I think it would be enough to add it to the dev dependencies as it is just used in a script. What do you think?

Uzlopak:
@p-kuen 

Please provide a PR to move c8 to devDependencies.

Eomm:
@fastify/collaborators Please - be mindful that we should put extra care when editing the package.json

simoneb:
I asked a review from Copilot to see if it would have caught it.

simoneb:
Looks like it didn't catch it 🤷

PR #581, opened 2025-05-19
update readme to use module imports

AlvesJorge:


mcollina (approved):
lgtm

Fdawgs (commented):
Tentative on this, majority of our docs are in CJS and the underlying code is in CJS.

Likewise, CJS is still the default for Node in that it will treat all `.js` files as cjs unless the type is explicitly set to module in package.json.

README.md
```diff
@@ -23,11 +23,12 @@ npm i @fastify/multipart
...
-const fastify = require('fastify')()
-const fs = require('node:fs')
-const { pipeline } = require('node:stream/promises')
+import fastify from "fastify"
+import fs from "node:fs"
+import { pipeline } from "node:stream";
+import multipart from "@fastify/multipart";
 
-fastify.register(require('@fastify/multipart'))
+fastify.register(multipart)
```
gurgunday, line 31: I think it's better to await this

AlvesJorge: Why? I want to understand for my own personal use as well.

jsumners:
> Tentative on this, majority of our docs are in CJS and the underlying code is in CJS.
> 
> Likewise, CJS is still the default for Node in that it will treat all `.js` files as cjs unless the type is explicitly set to module in package.json.

I agree with this. 

Also, each example, considering the diff only, seems to be completely stand alone prior to this proposed change. With this change, each example, subsequent to the first, references a missing variable.

AlvesJorge:
Thank you for sharing your perspective @jsumners, I didn't know this would be so polarizing!
The example on the readme over on [fastify/fastify](https://github.com/fastify/fastify) is also using ESM Imports, so it just made sense to me that the endorsed plugins would also follow suit :) 
As for the argument "the rest of our docs are in cjs", this would imply that the right thing to do would be for this PR to be extended to cover all existing docs, something that I don't have the time for.

I'm unsure what you mean by: 
> With this change, each example, subsequent to the first, references a missing variable.

What missing variable are you referring to?

AlvesJorge:
Thank you for sharing your perspective @jsumners, I didn't know this would be so polarizing!
The example on the readme over on [fastify/fastify](https://github.com/fastify/fastify) is also using ESM Imports, so it just made sense to me that the endorsed plugins would also follow suit :) 
As for the argument "the rest of our docs are in cjs", this would imply that the right thing to do would be for this PR to be extended to cover all existing docs, something that I don't have the time for.

I'm unsure what you mean by: 
> With this change, each example, subsequent to the first, references a missing variable.

What missing variable are you referring to?

is2ei:
> The example on the readme over on fastify/fastify is also using ESM Imports, so it just made sense to me that the endorsed plugins would also follow suit :)

The fastify/fastify repository’s README provides both ESM and CommonJS examples (please see attached).
If this change is intended to align with the README, it might be nice to include both ESM and CommonJS examples here as well.

https://github.com/fastify/fastify?tab=readme-ov-file#example

スクリーンショット 2025-08-23 0 52 11

PR #612, opened 2026-03-23
feat: return values from saveRequestFiles

### LINKED ISSUE #549: `saveRequestFiles` is consuming value fields and give no way to retreive them if there were no files
### Prerequisites

### Fastify version

3.20.x

### Plugin version

6.0.1

### Node.js version

21.7.1

### Operating system

Linux

Alpine 3.15 I think

### Description

I have a endpoint, where I process optional files. I really like the convenience of `saveRequestFiles`, but there is one problem - if there were no files - there is no way to access fields (I don't want to use `attachFieldsToBody` btw) - they are already consumed.

```javascript
    // this prints "start" and "end" only

    const files = await req.saveRequestFiles();
    // `files` is  an empty array
    console.log("start");
        console.log(part);
    }
    console.log("end");
```

I know I can still do stuff manually, but maybe there is a way to improve `saveRequestFiles` somehow? Even "dumb" stuff like attaching extra `fields` property directly to an array returned by `saveRequestFiles()` would be nice (that would be non-breaking change)

_No response_

### Expected Behavior

_No response_

mcollina:
Fixes #549.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,8 @@ This will store all files in the operating system's default directory for tempor
 
 ```js
 fastify.post('/upload/files', async function (req, reply) {
-  // stores files to tmp dir and return files
-  const files = await req.saveRequestFiles()
+  // stores files to tmp dir and returns files + parsed values
+  const { files, values } = await req.saveRequestFiles()
   files[0].type // "file"
   files[0].filepath
   files[0].fieldname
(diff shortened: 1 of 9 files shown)


jsumners (commented):
CI is red.

jsumners (approved):

gurgunday (approved):

Tony133 (approved):

PR #623, opened 2026-06-12
fix: expose InvalidJSONFieldError and NoFormData in multipartErrors

LeSingh1:
## What

Two error constructors defined inside the plugin were missing from the `fastify.multipartErrors` decorator:

- `InvalidJSONFieldError` (code `FST_INVALID_JSON_FIELD_ERROR`) — thrown when a multipart field has `Content-Type: application/json` but carries invalid JSON
- `NoFormData` (code `FST_NO_FORM_DATA`) — thrown when `globalThis.FormData` is unavailable

The TypeScript `MultipartErrors` interface was also missing `FileBufferNotFoundError`, `InvalidJSONFieldError`, and `NoFormData`, leaving the type out of sync with the runtime.

## Why it matters

Users who want to branch on the error type idiomatically do:

```js
} catch (err) {
  if (err instanceof fastify.multipartErrors.InvalidJSONFieldError) { ... }
}
```

With `InvalidJSONFieldError` absent from `multipartErrors`, that check throws a `TypeError: Right-hand side of 'instanceof' is not callable` instead of returning `false`, which is a silent breakage. The error *is* thrown by the plugin with the documented code, but there was no way to reach its constructor through the public API.

## What changed

- `index.js`: add `InvalidJSONFieldError` and `NoFormData` to the `fastify.decorate('multipartErrors', …)` call
- `types/index.d.ts`: add `FileBufferNotFoundError`, `InvalidJSONFieldError`, and `NoFormData` to `MultipartErrors`
- `test/multipart-json.test.js`:
  - Fix `t.plan(2)` → `t.plan(3)` in two existing tests that were silently miscounted (the instanceof check was throwing a TypeError, skipping one assertion, and the wrong plan count went unnoticed)
  - Add a unit test confirming `multipartErrors.InvalidJSONFieldError` is a function and that an error thrown by the plugin is an instance of it

All tests pass (`npm test`) including TypeScript type checks.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -62,7 +62,10 @@ interface MultipartErrors {
   PrototypeViolationError: FastifyErrorConstructor;
   InvalidMultipartContentTypeError: FastifyErrorConstructor;
   RequestFileTooLargeError: FastifyErrorConstructor;
+  FileBufferNotFoundError: FastifyErrorConstructor;
   PrematureCloseError: FastifyErrorConstructor;
+  InvalidJSONFieldError: FastifyErrorConstructor;
+  NoFormData: FastifyErrorConstructor;
 }
 
 declare namespace fastifyMultipart {
--- a/index.js
+++ b/index.js
@@ -181,7 +181,9 @@ function fastifyMultipart (fastify, options, done) {
     InvalidMultipartContentTypeError,
     RequestFileTooLargeError,
     FileBufferNotFoundError,
-    PrematureCloseError
+    PrematureCloseError,
+    InvalidJSONFieldError,
+    NoFormData
   })
 
   fastify.addContentTypeParser('multipart/form-data', setMultipart)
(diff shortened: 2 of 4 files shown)


Eomm (commented):
Would you mind to rebase this PR?

types/index.d.ts
```diff
@@ -62,7 +62,10 @@ interface MultipartErrors {
   PrototypeViolationError: FastifyErrorConstructor;
   InvalidMultipartContentTypeError: FastifyErrorConstructor;
   RequestFileTooLargeError: FastifyErrorConstructor;
+  FileBufferNotFoundError: FastifyErrorConstructor;
   PrematureCloseError: FastifyErrorConstructor;
+  InvalidJSONFieldError: FastifyErrorConstructor;
+  NoFormData: FastifyErrorConstructor;
```
mcollina, line 68: Can you add these to the type tests?

PR #624, opened 2026-06-17
fix(types): add optional value to MultipartFile for onFile

### LINKED ISSUE #574: Type value does not exist on MultipartFile when using onFile
### Prerequisites

- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.2.1

### Plugin version

9.0.3

### Node.js version

23.7.0

### Operating system

Windows

11

### Description

When using `onFile`, the part does not have typings for `value`

So this code from the readme is not valid. As such I am providing this github repo as the reproducible example:

```js
async function onFile(part) {
  const buff = await part.toBuffer()
  const decoded = Buffer.from(buff.toString(), 'base64').toString()
  part.value = decoded // set `part.value` to specify the request body value
}

fastify.register(require('@fastify/multipart'), { attachFieldsToBody: 'keyValues', onFile })

fastify.post('/upload/files', async function (req, reply) {
  const uploadValue = req.body.upload // access file as base64 string
  const fooValue = req.body.foo       // other fields
})
```

https://github.com/fastify/fastify-multipart

### Expected Behavior

Typings for `value` on `MultipartFile`

imharjot:
The README documents setting `part.value` inside an `onFile` handler when using
`attachFieldsToBody: 'keyValues'`, but `MultipartFile` has no `value` property, so that
example fails to typecheck with TS2339 (`Property 'value' does not exist on type 'MultipartFile'`).

Adds an optional `value?: unknown` to `MultipartFile`. `unknown` (not `any`) keeps it
type-safe and matches the existing `Record<string, unknown>` usage in this file; it's
optional because a file part only gets a `value` when one is assigned in `onFile`. Kept
`MultipartFile` non-generic to avoid rippling generics through `Multipart`/`MultipartFields`.

The existing type test asserted `req.body.file` had no `value`, so that assertion is flipped
and the `onFile` test handler now assigns `part.value` to cover the regression.

Fixes #574

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -81,6 +81,11 @@ declare namespace fastifyMultipart {
 
   export interface MultipartFile {
     type: 'file';
+    /**
+     * A value can be assigned to a file part inside an `onFile` handler when
+     * using `attachFieldsToBody: 'keyValues'` to specify the decoded body value.
+     */
+    value?: unknown;
     toBuffer: () => Promise<Buffer>;
     file: BusboyFileStream;
     fieldname: string;
--- a/types/index.tst.ts
+++ b/types/index.tst.ts
@@ -20,6 +20,7 @@ const runServer = async () => {
       parts: 500
     },
     onFile: (part: MultipartFile) => {
+      part.value = 'decoded'
       console.log(part)
     }
   })
@@ -73,7 +74,7 @@ const runServer = async () => {
     reply.send()
 
     expect(req.body.file.file).type.toBe<BusboyFileStream>()
-    expect(req.body.file).type.not.toHaveProperty('value')
+    expect(req.body.file).type.toHaveProperty('value')
   })
 
   // busboy


Tony133 (approved):

is2ei (commented):
I think the issue is a bit more complicated.

It looks like assigning the `value` field to the `part` object in the `onFile` function is a bit of a hack.
This works because the `if` statement uses `field.value` when it is not `undefined`.

https://github.com/fastify/fastify-multipart/blob/main/index.js#L100-L103
```js
            if (field.value !== undefined) {
              body[key] = field.value
            } else if (field._buf) {
              body[key] = field._buf
```

However, the `value` property is intended for objects whose `type` is `field`.

https://github.com/fastify/fastify-multipart/blob/main/index.js#L338-L347

```js
      const value = {
        type: 'field',
        fieldname: name,
        mimetype: contentType,
        encoding,
        value: fieldValue,
        fieldnameTruncated,
        valueTruncated,
        fields: body
      }
```

On the other hand, the `_buf` property is intended for objects whose `type` is `file`.

https://github.com/fastify/fastify-multipart/blob/main/index.js#L373-L382

```js
      const value = {
        type: 'file',
        fieldname: name,
        filename,
        encoding,
        mimetype,
        file,
        fields: body,
        _buf: null,
        async toBuffer () {
```

PR #631, opened 2026-07-06
fix: do not end parts() iteration while busboy still holds undelivered parts

### LINKED ISSUE #630: parts() can end iteration while parts remain undelivered (trailing file parts silently lost on reused connections)
## Summary

`request.parts()` can end iteration while busboy still has parts to deliver. In our production app the second multipart POST on a reused keep-alive connection reliably lost its trailing file part and all trailing fields: the async iterator simply ended after the first file, with no error thrown. Swapping the iterator for a busboy-driven one that ends only on busboy's own `finish` eliminated the problem completely in the same environment.

Reproduced on `@fastify/multipart` 9.4.0, and the relevant code is unchanged in 10.0.0.

## Environment

- fastify 5.9.0
- @fastify/multipart 9.4.0 (code inspected in 10.0.0 as well)
- Node 22 (Debian bookworm, in Docker; clients reached it through Docker Desktop port forwarding)
- Clients that triggered it: Chrome (XHR + FormData) and curl with connection reuse (`curl url -F ... --next url -F ...`)

## Observed behavior

Handler shape (simplified; the real one streams each file part to object storage):

```js
app.post("/upload", async (req, reply) => {
  for await (const part of req.parts()) {
    if (part.type === "file") {
      if (part.fieldname === "mask") maskKey = await save(part);
      else imageKey = await save(part);
    } else {
      // fields: clientJobId, format, quality
    }
  }
  if (!maskKey) return reply.status(400).send({ error: "no mask" });
});
```

Client sends `file` (87 KB), `mask` (6 KB), then three small fields, with correct `Content-Length`, verified at the XHR layer (both File entries present with nonzero sizes) and by a curl repro.

- First POST on a fresh connection: all five parts arrive. 202.
- Second POST on the same keep-alive connection: the iterator yields only the first file part, then ends cleanly. The stored first file is byte-complete (all 89084 bytes), no error is thrown, the loop just finishes. Our mask check then 400s. Response time ~4 ms, so the consumer was not slow in the failing case.
- Alternates reliably: fresh connection OK, reused connection drops, on both Chrome and curl.

curl repro against the affected app (two identical requests, one connection):

```
req1: 202          (uploads/<job1>/ contains file + mask)
req2: 400 no mask  (uploads/<job2>/ contains ONLY the first file, byte-complete)
```

## Code paths that can end iteration early

Two places in `index.js` end the parts iterator while busboy may still hold undelivered parts:

1. `cleanup` is bound to `request.on('close', ...)` (and `error`). `cleanup` pushes the end marker into the same single queue the parts flow through (`ch(err || lastError || null)`). `IncomingMessage` emits `close` once the message completes, which with a fully buffered body (fast client, reused connection, userspace proxies re-segmenting the stream) can happen while the consumer is still awaiting an earlier part's storage write. Any part busboy emits after that lands behind the end marker and is never yielded; `parts()` resolves `null` first and iteration stops.

2. Busboy's `Unexpected end of multipart data` error is translated into a clean end of iteration instead of an error (`resolve(null)` at index.js:245 in 9.4.0). If the source ends unexpectedly mid-stream for any reason, the consumer sees a normal, complete-looking iteration that just happens to be missing the trailing parts. Nothing distinguishes "done" from "truncated".

We were not able to reduce the environment-specific timing to a single-file repro on loopback (the queue ordering comes out safe there because busboy's final `_write` acks only after synchronously emitting the buffered parts, so `end`/`close` queue behind them). Through Docker Desktop's port forwarding the failure is deterministic on the second request per connection. We are happy to run instrumented builds or bisect further if that helps.

## Workaround that eliminates it

Driving busboy directly and ending iteration only on busboy's `finish` (treating request `error` and busboy `error` as throws, and never treating request `close` as an end signal) fixed it outright in the same environment where the stock iterator failed deterministically:

```js
export async function* multipartParts(request) {
  const raw = request.raw;
  const bb = new Busboy({ headers: raw.headers });
  const queue = [];
  let wake = null;
  const push = (v) => { queue.push(v); wake?.(); wake = null; };

  bb.on("file", (fieldname, stream, filename, encoding, mimetype) =>
    push({ type: "file", fieldname, filename, encoding, mimetype, file: stream }));
  bb.on("field", (fieldname, value) => push({ type: "field", fieldname, value }));
  bb.on("error", (err) => push(err instanceof Error ? err : new Error(String(err))));
  bb.on("finish", () => push(DONE));
  raw.on("error", (err) => push(err));
  raw.pipe(bb);

  try {
    while (true) {
      if (queue.length === 0) await new Promise((r) => { wake = r; });
      const v = queue.shift();
      if (v === undefined) continue;
      if (v === DONE) return;
      if (v instanceof Error) throw v;
      yield v;
    }
  } finally {
    raw.unpipe(bb);
    bb.removeAllListeners();
  }
}
```

## Suggested direction

- End iteration on busboy's `finish` rather than on the request stream's `close`; `close` on a completed message is not a statement about undelivered buffered parts.
- Surface `Unexpected end of multipart data` as an error to the consumer instead of a clean end, so truncation is never silent. Consumers validating "did I get all expected parts" currently cannot tell the difference.

Thanks for the plugin. Happy to provide more forensics, run patches against the environment that reproduces this, or open a PR along the lines above if you want to go that way.

mcollina:
## Summary

Fixes #630.

`request.on('close')` fires once the message body has fully drained into busboy's writable buffer. Busboy defers its write ack when a file stream backs up, so on a fast/re-segmented connection the request can complete while busboy still holds parsed-but-undelivered parts. The `cleanup` bound to `close` pushed the end marker into the same queue the parts flow through, and every part emitted afterwards landed behind it — trailing file parts and fields were silently lost, exactly as reported.

## Changes

- **End iteration on busboy's own end events.** `request.on('close')` only terminates the iteration when the message did **not** complete (`readableEnded === false`, i.e. a client abort), and then with a new `PrematureCloseError` (`FST_MP_PREMATURE_CLOSE`, exposed via `fastify.multipartErrors` and the TypeScript types) instead of a silent clean end.
- **Surface `Unexpected end of multipart data` as an error.** The suppression came in mechanically with the busboy v2.1.0 upgrade (#506); it made truncated uploads indistinguishable from complete ones. A fully empty body (zero bytes received) is still treated as an empty form, preserving the behavior pinned by `multipart-empty-body.test.js`.
- **End iteration when the consumer destroys a file stream without reading it to the end.** Busboy never emits `finish` in that case; previously the `close`-based cleanup masked this, and without a guard `parts()` would hang forever (caught by the existing *should not freeze when error is thrown during processing* test).

## Reproduction / tests

Written TDD-style in `test/fix-630.test.js` (red on `main`, green here):

1. **Trailing parts survive early request `close`** — slow consumer, 200 KB file + 6 KB mask + three fields. A large busboy `highWaterMark` makes the failing event ordering deterministic on loopback (the same ordering the issue reproduced through Docker Desktop's port forwarding with default settings); on `main` the handler sees only the first file.
2. **Truncated multipart data rejects** with `Unexpected end of multipart data` instead of ending cleanly.
3. **Client abort mid-upload rejects the iterator** — guards that dropping the `close`-based cleanup doesn't reintroduce hangs on aborts.

## Behavioral note

Consumers that previously saw a clean-looking, silently-truncated iteration on truncated or aborted uploads now get an error — that is the point of the fix, but it may warrant a semver-major release.

## Checklist

- [x] run `npm run test` — 90/90 unit tests, 100 % coverage, tstyche 34/34
- [x] documentation is changed or added — no doc changes needed (error is exported via existing `multipartErrors` mechanism)
- [x] commit message follows commit guidelines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ const PrototypeViolationError = createError('FST_PROTO_VIOLATION', 'prototype pr
 const InvalidMultipartContentTypeError = createError('FST_INVALID_MULTIPART_CONTENT_TYPE', 'the request is not multipart', 406)
 const InvalidJSONFieldError = createError('FST_INVALID_JSON_FIELD_ERROR', 'a request field is not a valid JSON as declared by its Content-Type', 406)
 const FileBufferNotFoundError = createError('FST_FILE_BUFFER_NOT_FOUND', 'the file buffer was not found', 500)
+const PrematureCloseError = createError('FST_MP_PREMATURE_CLOSE', 'the request was closed before the multipart data was fully parsed', 400)
 const NoFormData = createError('FST_NO_FORM_DATA', 'FormData is not available', 500)
 
 function setMultipart (req, _payload, done) {
@@ -179,7 +180,8 @@ function fastifyMultipart (fastify, options, done) {
     PrototypeViolationError,
     InvalidMultipartContentTypeError,
     RequestFileTooLargeError,
-    FileBufferNotFoundError
+    FileBufferNotFoundError,
+    PrematureCloseError
   })
 
   fastify.addContentTypeParser('multipart/form-data', setMultipart)
@@ -243,16 +245,8 @@ function fastifyMultipart (fastify, options, done) {
     const parts = () => {
       return new Promise((resolve, reject) => {
         handle((val) => {
-          if (val instanceof Error) {
-            if (val.message === 'Unexpected end of multipart data') {
-              // Stop parsing without throwing an error
-              resolve(null)
-            } else {
-              reject(val)
-            }
-          } else {
-            resolve(val)
-          }
+          if (val instanceof Error) return reject(val)
+          resolve(val)
         })
       })
     }
@@ -260,6 +254,7 @@ function fastifyMultipart (fastify, options, done) {
     const body = {}
     let lastError = null
     let currentFile = null
+    let sawData = false
     const request = this.raw
     const busboyOptions = deepmergeAll(
       { headers: request.headers },
@@ -270,7 +265,7 @@ function fastifyMultipart (fastify, options, done) {
     this.log.trace({ busboyOptions }, 'Providing options to busboy')
     const bb = busboy(busboyOptions)
 
-    request.on('close', cleanup)
+    request.on('close', onRequestClose)
     request.on('error', cleanup)
 
     bb
@@ -279,7 +274,7 @@ function fastifyMultipart (fastify, options, done) {
       .on('end', cleanup)
       .on('finish', cleanup)
       .on('close', cleanup)
-      .on('error', cleanup)
+      .on('error', onBusboyError)
 
     bb.on('partsLimit', function () {
       const err = new PartsLimitError()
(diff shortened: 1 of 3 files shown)


gurgunday (approved):
Lgtm

Issue #620, opened 2026-05-13
Compatibility with preParsing / onRequest hooks


tmcw:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

As far as I can tell, this plugin reads from `request.raw`:

https://github.com/fastify/fastify-multipart/blob/60e1b5ca627602c7e1fbf12209ec2af0828b8dc0/index.js#L263

In my usecase, which is verifying webhook signatures from Sendgrid, I'll need to have the raw binary body so that the webhook can be verified, but then parse the body as multipart. Getting both of these right does not appear to be possible with how fastify-multipart is set up right now.

If I use `preParsing` and read the request stream to verify it, then this plugin looks and finds the request stream empty and closed by the time it gets to the verification hook.

### Motivation

Making this plugin more intuitive to use with other hooks - it did not occur to me for a long time that my preParsing hook which followed documented patterns would not work because this plugin was reading from req.raw instead of reading the payload stream.

PR #634, opened 2026-07-22
feat: allow manually registering the attachFieldsToBody preValidation hook

### LINKED ISSUE #599: Support manual hook injection
### Prerequisites

- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I have multiple hooks during the preValidation step including hooks that modify the content or run a signature check on the content. Because the hook is registered at the instance boot and not at the route level i cannot decide on the hooks order.

https://github.com/fastify/fastify-multipart/blob/83aea9789d45a6222041a4fa33fd9535a179f4b9/index.js#L70

Would it be possible to put the hook in a dedicated exported function and add a new parameter to allow to manually load the hook via the function?

### Motivation

_No response_

### Example

_No response_

Truta446:
## Summary

Fixes #599.

When `attachFieldsToBody` is enabled, the plugin registers its own
`preValidation` hook at the instance level. Because instance-level hooks
always run before route-level ones (and in registration order relative to
other instance-level hooks in the same scope), users had no way to control
this hook's position relative to their own `preValidation` hooks — e.g. a
hook that verifies a signature on the raw body before any field parsing
happens.

This PR extracts the hook body into a method decorated on the request,
`request.attachFieldsToBody()`, and adds a new `attachFieldsToBodyHook`
option (default `true`, matching current behavior). When set to `false`,
the plugin skips automatic registration and the user calls
`request.attachFieldsToBody()` themselves, from whichever hook and in
whatever order they choose:

```js
fastify.register(require('@fastify/multipart'), { attachFieldsToBody: true, attachFieldsToBodyHook: false })

fastify.addHook('preValidation', async function (req) {
  await verifySignature(req) // runs first, now controllable

  if (req.isMultipart()) {
    await req.attachFieldsToBody()
  }
})
```

No behavior change when `attachFieldsToBodyHook` is left unset.

## Test plan

- Added a test in `test/multipart-attach-body.test.js` reproducing the
  issue's scenario: a custom `preValidation` hook that must run before
  field parsing, using `attachFieldsToBodyHook: false` to control ordering.
- `npm run test:unit` — 91/91 passing, 100% coverage.
- `npm run test:typescript` — passing.
- `npm run lint` — clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,20 @@ fastify.post('/upload/files', async function (req, reply) {
 You can only use the `toBuffer` method to read the content.
 If you try to read from a stream and pipe to a new file, you will obtain an empty new file.
 
+When `attachFieldsToBody` is enabled, the plugin registers its own `preValidation` hook to do the parsing. Because instance-level hooks always run before route-level ones, this hook always runs first among your other `preValidation` hooks, and you can't reorder it relative to them. If you need to run something before it (e.g. verifying a signature on the raw body), set `attachFieldsToBodyHook: false` to skip automatic registration, then call `request.attachFieldsToBody()` yourself from a hook you control:
+
+```js
+fastify.register(require('@fastify/multipart'), { attachFieldsToBody: true, attachFieldsToBodyHook: false })
+
+fastify.addHook('preValidation', async function (req) {
+  await verifySignature(req) // runs first
+
+  if (req.isMultipart()) {
+    await req.attachFieldsToBody() // parses fields into req.body
+  }
+})
+```
+
 ## JSON Schema body validation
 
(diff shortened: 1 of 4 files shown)


climba03003 (changes requested):
Verify signature means it needs to read the payload.
When the payload is consumed by the user, `attachFieldsToBody` can't do anything later.

Truta446:
You're absolutely right, verifying a signature means reading the payload, and once it's read attachFieldsToBody() has nothing left to parse. That use case isn't what this option solves; it belongs in a preParsing hook that tees the stream (which is what #620 enables). I've dropped the signature-verification framing from this PR: the README example now shows a correct use case (rejecting early, before the plugin buffers the upload, using headers/token rather than the body), plus an explicit note that attachFieldsToBody() consumes the stream and that payload-reading concerns like signature verification should tee the stream in preParsing instead. The test was updated to match.

PR #645, opened 2026-09-21
docs: clarify multipart bodyLimit is not an aggregate limit

mcollina:
## Summary

Document that `@fastify/multipart` does not enforce Fastify's aggregate `bodyLimit` for multipart requests.

Multipart payloads are parsed by a dedicated streaming parser and bypass Fastify's `bodyLimit` enforcement. `bodyLimit` is only used to compute the default `fileSize`; it is **not** enforced as an aggregate limit across the whole request. This is intentional (streaming) but was undocumented, and the old README text (`fileSize` defaulting to _1048576_) was misleading because that value only holds when `bodyLimit` is the Fastify default.

## Changes
- README: replace the stale "_1000_ and _1048576_" default note, and add an explicit caveat that `bodyLimit` is not an aggregate limit and is not applied at the route level.
- types: note that `limits.fileSize` defaults to the Fastify `bodyLimit`.

This is a documentation-only change, no runtime behavior.

Ref: GHSA-3pm6-wjfr-3v47


Tony133:
CI all green 🙌 ✅

Tony133 (approved):
lgtm

